PR #1288, opened 2014-09-19
fixes bug where large amount of uploads races ahead of UI

craigrbruce:
We have a bug where, when a large amount of files being uploaded (go to a folder with lots of files, select all) the update goes to the upload tray but the element doesn't exist there yet so the prepend function in templating.js throws an exception and all subsequent files fail to submit to the uploader.

    qq(parentEl).children()[index].nextSibling;

can not call nextSibling of undefined.

This fix is just a simple guard to avoid this and allow all files to be submitted.

Tested on Chrome, Firefox and Safari

### DIFF
--- a/client/js/templating.js
+++ b/client/js/templating.js
@@ -213,8 +213,8 @@ qq.Templating = function(spec) {
             beforeEl = parentEl.firstChild;
 
         if (index > 0) {
-            beforeEl = qq(parentEl).children()[index].nextSibling;
-
+            var child = qq(parentEl).children()[index];
+            child && (beforeEl = child.nextSibling);
         }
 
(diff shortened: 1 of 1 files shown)


rnicholus:
Just out of curiosity, are you still seeing this issue as of Fine Uploader 5.4.1?

craigrbruce:
Hi Ray

I have since changed companies and am no longer using Fine Uploader day to day. I don't remember seeing this issue with the latest version we were using, though I can not 100% verify this.

Good luck.

rnicholus:
Thanks for the update. I'm going to close this then, but will consider re-opening if anyone is still seeing this on a recent version of the library.

PR #1345, opened 2015-01-08
Reject all files which fail validation

ahorner:
Currently, when multiple files are submitted, some or all of which fail 
validation, the files which fail are left lingering in SUBMITTING status
instead of being moved to the REJECTED state. This has all sorts of 
nasty side effects, including preventing the `allComplete` callback from
ever firing once the valid files in the batch have been successfully 
uploaded.

The solution is simply to ensure that the passed index is used when 
grabbing the file from `items` in the validation failure callback, so 
the file which fails validation is properly rejected.

This is a small change, but I tested against a few browsers to make
sure I didn't break anything else in the process of fixing this issue.


rnicholus:
Thanks for catching this.  Can you please [sign the CLA](https://www.clahub.com/agreements/FineUploader/fine-uploader)?

ahorner:
Already done.

rnicholus:
Ah, so you have.  I'll take a closer look at this tomorrow.

rnicholus:
While writing tests for this issue, I believe I may have encountered another somewhat related issue, which I will track down as well.  Once that is all straightened out and tests are passing, I'll probably release all of this as a hotfix: 5.1.3.

rnicholus:
This has been fixed and released as Fine Uploader 5.1.3.

ahorner:
Thanks for looking into it.

Issue #804, opened 2013-04-22
Display 0.1GB as 100MB


rdepena:
100 megabytes (1024*1024 \* 100) passed as sizeLimit will display as 0.1GB instead of 100MB. on the sizeError. All versions.

PR #1356, opened 2015-01-21
Fix file size approximation.

menthol:
Fix filesize displayed to users

1 KB = 1024 B
1MB = 1024 KB


rnicholus:
This is not broken.  Please see the discussion in #804 for more information.

PR #1360, opened 2015-02-07
Corrected typo

thanosp:
- Corrected typo: `requsterOptions` should be `requesterOptions`


rnicholus:
We generally don't accept pull requests against master, unless we are dealing with a hotfix.  This falls under the category of refactor as this is just a rename of an internal variable.  Thanks anyway though.

Issue #1212, opened 2014-05-22
3 - Azure: set Content-Type, Content-Disposition, etc blob headers


AnalogJ:
I'm sorry if this is the wrong place to ask this, but I wasn't able to find anything in the docs or on SO. 

With the Azure fineuploader jquery plugin, is it possible to set the content-type or content-disposition headers for a file that is being uploaded? The `params` object seems to have all its headers name-spaced with `x-ms-meta`.

Korijn:
Imho, not prefixing keys passed into setParams should be fine? Overwrite an existing header: no problem... it's apparently the developer's choice.

Korijn:
Or provide an option to toggle auto prefixing of params?

rnicholus:
I think params should always be prefixed. I can see allowing creation of any headers via a `setCustomHeaders` API method for the Azure module though.

Korijn:
:+1:

Korijn:
I went ahead and created a pull request, as I'm eager to start using the validation mechanism provided by Content-MD5.

PR #1404, opened 2015-04-30
Add unprefixed Azure request headers support

Korijn:
Adds support for a predefined list of headers which are appropriate for
Azure without the x-ms-meta- prefix.

Resolves issue #1212 and is based on similar pull request #1258.

### DIFF
--- a/client/js/azure/util.js
+++ b/client/js/azure/util.js
@@ -6,11 +6,41 @@ qq.azure.util = qq.azure.util || (function() {
     return {
         AZURE_PARAM_PREFIX: "x-ms-meta-",
 
+        /** Create Prefixed request headers which are appropriate for Azure.
+         *
+         * If the request header is appropriate for Azure (e.g. Cache-Control) then pass
+         * it along without a metadata prefix. For all other request header parameter names,
+         * apply qq.azure.util.AZURE_PARAM_PREFIX before the name.
+         *
+         * @param name Name of the Request Header parameter to construct a (possibly) prefixed name.
+         * @returns {String} A valid Request Header parameter name.
+         */
+        _getPrefixedParamName: function (name) {
+            switch (name)
+            {
+                //
+                // Valid request headers (not sent by fine-uploader) which should be returned as-is (case-sensitive)
+                // see: https://msdn.microsoft.com/en-us/library/azure/dd179451.aspx
+                //
+                case "Cache-Control":
+                case "Content-Disposition":
+                case "Content-Encoding":
+                case "Content-MD5":
+                case "x-ms-blob-content-encoding":
+                case "x-ms-blob-content-disposition":
+                case "x-ms-blob-content-md5":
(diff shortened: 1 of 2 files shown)


Korijn:
I tried to sign the CLA but the clahub just barfs "something went wrong" after I fill in the form and submit.

rnicholus:
Have you built and tested to make sure this functions exactly as you expect it to?

Korijn:
Admittedly, I was in a rush. I took the time to carefully test it now though. The `Content-MD5` request header can be set using `setParams({ 'Content-MD5', b64_hash }, id)`. Tested in IE11, Firefox 37 and Chrome 42.

test

rnicholus:
Thanks, I'll merge this into develop to be part of the next scheduled release just as soon as I get some time to write a unit test or two, schedule permitting. If you're able to write a test instead, I'll be able to guarantee that this makes it into the next release.

Korijn:
How's this?

rnicholus:
@Korijn Looks good now! Can you take a peek at the [contributor agreement](https://github.com/FineUploader/fine-uploader/blob/master/CONTRIBUTING.md#contributor-agreement) and indicate here if you agree to its terms?

Korijn:
Agreed. :+1:

client/js/azure/util.js
```diff
@@ -6,11 +6,55 @@ qq.azure.util = qq.azure.util || (function() {
...
+        },
+
         getParamsAsHeaders: function(params) {
             var headers = {};
 
             qq.each(params, function(name, val) {
-                var headerName = qq.azure.util.AZURE_PARAM_PREFIX + name;
+                var headerName = qq.azure.util._getPrefixedParamName(name);
 
                 if (qq.isFunction(val)) {
```
Korijn: I guess this if-else-tree might cause some unexpected behavior... what if the value is a function ánd the parameter name matches an azure parameter? I'll improve upon it next Monday if you agree.

Korijn: I'm a bit short on time for this final adjustment, can we postpone it till after release?

rnicholus: The next scheduled release is probably at least a couple months away. Either way this change can wait, so, not a problem.

Korijn:
I think the PR is done now.

I could include some more [headers from the Microsoft docs](https://msdn.microsoft.com/en-us/library/azure/dd179451.aspx) but I think you already handle a bunch of them specifically, so I kept the PR restricted to the ones I could test in my setup.

Korijn:
I just rebased it on the latest develop (I think) so should be ready to merge if the travis build passes.

rnicholus:
Thanks, this looks good. I may end up rolling the hotfix/5.2.3 branch into develop and release this, along with everything currently tagged as 5.2.3 as part of a new 5.3 release (pushing all existing 5.3 stuff to 5.4).

PR #1423, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


rnicholus:
Why?

kkirsche:
Because it's not used and has never been used. It's completely pointless to include.

rnicholus:
We don't directly publish to bower, so one could argue that the entire bower.json file is irrelevant. There's nothing special about the version property. Not sure this merits a change to master either.

kkirsche:
To each their own. I strongly disagree with that view, but whatever

rnicholus:
Ah, I just noticed the link in your description. Regardless, all pull requests must be against develop? As mentioned in the contributing doc. I'll remove this property in the next release though.

kkirsche:
This is literally going to every single person who has this in their bower.json. I'm not bothering with reading each contrib doc. I do apologize for not following it, but too many to do.

rnicholus:
I understand. Thanks for the heads up though, and like I said, I'll remove this in the near future.

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #1458, opened 2015-08-18
Hide retry button when status changes to UPLOAD_RETRYING instead of directly in hardcoded scenarios.

### LINKED ISSUE #1455: Retry button remains visible when initiating a retry programmatically
For example, when applying the following mechanism to arrange for endless retries, the Retry button remains visible, even after the upload succeeds.

```javascript
uploader = new qq.azure.FineUploader({
    callbacks: {
        onStatusChange: function (id, oldStatus, newStatus) {
            if (newStatus === qq.status.UPLOAD_FAILED) {
                uploader.retry(id);
```

Korijn:
This PR fixes #1455.

### DIFF
--- a/client/js/uploader.api.js
+++ b/client/js/uploader.api.js
@@ -171,7 +171,6 @@
                 },
 
                 onRetry: function(fileId) {
-                    qq(self._templating.getFileContainer(fileId)).removeClass(self._classes.retryable);
                     self.retry(fileId);
                 },
 
--- a/client/js/uploader.api.js
+++ b/client/js/uploader.api.js
@@ -172,7 +172,6 @@
 
                 onRetry: function(fileId) {
                     qq(self._templating.getFileContainer(fileId)).removeClass(self._classes.retryable);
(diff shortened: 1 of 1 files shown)


client/js/uploader.api.js
```diff
@@ -172,7 +172,6 @@
 
                 onRetry: function(fileId) {
                     qq(self._templating.getFileContainer(fileId)).removeClass(self._classes.retryable);
```
rnicholus: @Korijn This line should be removed as well. Otherwise, all looks good!

Korijn:
Should be good now?

rnicholus:
Yep, looks good. Were you also able to verify that this works as expected for you?

Korijn:
Yes, I'm already using a temporary build sucessfully in production (we really needed the feature).

rnicholus:
Great. I'll push this out as a hotfix release very soon. It will be 5.3.1.

Korijn:
Awesome, thanks for the help.

Issue #1479, opened 2015-10-08
Blobs uploaded via Azure module in uuid mode get .undefined extension if original filename has no extension


Korijn:
Blobs uploaded in uuid mode get .undefined extension if original filename has no extension. PR incoming.

rnicholus:
Can you describe the steps needed to reproduce?

Korijn:
When you use the Azure FineUploader, by default it will upload files using uuid's as the blob name. If the file you upload has no extension, FineUploader appends ".undefined" to the uuid.

Also see the PR I submitted, it's pretty obvious.

Korijn:
I just confirmed my custom build with the PR included as working on our testing servers, as we needed to get this live quickly.

rnicholus:
Ah, that's the critical piece that is missing from your PR and issue - Azure. I'll update both cases.

Korijn:
Yeah, sorry about that. It's so natural for me that I forgot to mention it. :)

rnicholus:
No problem. I'll probably release this as part of 5.4 instead of a hotfix release. Use of Azure, compared to S3, is _very_ small, so this likely doesn't affect many users.

Korijn:
Sure, we'll merge upstream into our custom builds until it's included. Thanks!

PR #1480, opened 2015-10-08
Fix for #1479: Blobs uploaded via Azure module with uuid mode get .undefined extension if original filename has no extension

Korijn:
Fix for #1479: Blobs with uuid mode get .undefined extension if original filename has no extension

### DIFF
--- a/client/js/azure/uploader.basic.js
+++ b/client/js/azure/uploader.basic.js
@@ -100,12 +100,12 @@
                 blobNameOptionValue = this._options.blobProperties.name,
                 uuid = this.getUuid(id),
                 filename = this.getName(id),
-                fileExtension = qq.getExtension(filename);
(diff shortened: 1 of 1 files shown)


rnicholus:
I forgot to ask you this in previous PRs, but do you agree to give up any future claims you may have to the code in this PR, since it will become part of Fine Uploader for which Widen holds a copyright?

Korijn:
I think I had to sign something when I first sent you a PR? But the form was broken, so I agreed textually in the PR discussion.

So, again, agreed!

Korijn:
See: https://github.com/FineUploader/fine-uploader/pull/1404#issuecomment-100174281

rnicholus:
Thanks for the fix. I just merged into develop and added a unit test to cover this.

PR #1508, opened 2015-12-15
Docs/few corrections

skvale:


rnicholus:
Thanks for these fixes Sam, and welcome to the Widen team! 

I'll update master as well.

PR #1516, opened 2016-01-12
Allow access to chunk data.

ericmackrodt:
As an use case in my company, we need to be able to encrypt each chunk separately when uploading a file. In order to do that, we needed access to the blobs of the chunks, so we created an event to support this feature.

I believe that it's an important feature to have in the library to allow more flexibility and I've seen other people on the internet who needed to do something similar. Here's an example:

http://stackoverflow.com/questions/31051664/how-to-modify-current-chunk-data-during-upload-process-with-fineuploader

I believe that, if the developer use it the wrong way and breaks their upload flow, it's entirely their responsibility.
Of course, there's a check to see if the content size of the chunk changed (except for the last one).

I hope you accept this Pull Request and maybe even improve on this feature in the future. I'm sure it will be helpful for many people.

### DIFF
--- a/client/js/traditional/traditional.xhr.upload.handler.js
+++ b/client/js/traditional/traditional.xhr.upload.handler.js
@@ -21,11 +21,11 @@ qq.traditional.XhrUploadHandler = function(spec, proxy) {
             var size = getSize(id),
                 name = getName(id);
 
-            params[spec.chunking.paramNames.partIndex] = chunkData.part;
-            params[spec.chunking.paramNames.partByteOffset] = chunkData.start;
-            params[spec.chunking.paramNames.chunkSize] = chunkData.size;
-            params[spec.chunking.paramNames.totalParts] = chunkData.count;
-            params[spec.totalFileSizeName] = size;
+            params[spec.chunking.paramNames.partIndex] = params.partIndex || chunkData.part;
+            params[spec.chunking.paramNames.partByteOffset] = params.partByteOffset || chunkData.start;
+            params[spec.chunking.paramNames.chunkSize] = params.chunkSize || chunkData.size;
+            params[spec.chunking.paramNames.totalParts] = params.totalParts || chunkData.count;
+            params[spec.totalFileSizeName] = params.totalFileSize || size;
 
             /**
              * When a Blob is sent in a multipart request, the filename value in the content-disposition header is either "blob"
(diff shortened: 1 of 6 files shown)


rnicholus:
Instead of opening a new PR each time, you can simply push fixes to your branch.

ericmackrodt:
Oh, didn't know that. Sorry.

rnicholus:
There are a _lot_ of changes here, many of which I don't agree with. Allowing the chunk to be potentially modified (made smaller or larger) adds quite a bit of complexity to the library, which itself is already more complex than I'd like it to be. However, I think this can still be accomplished for traditional endpoints by extending the `onUploadChunk` event/callback. Currently, Fine Uploader passes metadata about the chunk to the callback handler, just before it is uploaded. The code could be modified to include the chunk `Blob` as well. Furthermore, internal code that handles `onUploadChunk` calls could be modified to support a promissory return type that, when resolved, includes the new chunk `Blob`. I don't think this will be a terribly difficult modification to make. 

I'm going to close this pull request, but if you'd like me to consider this for a future release with my proposed solution, please open up a feature request, or, better yet, a pull request with minimal changes that implement my proposal.

ericmackrodt:
What you suggested would be ideal, the problem is that it wouldn't support asynchronous code within `onUploadChunk`. That's what I need to support since each of the chunks I'm sending to the server are being encrypted using a WebWorker and there's no way to wait for the async work to be done within `onUploadChunk` unless I add a callback to it the same way I did with the other event. But I wanted to keep the event that already existed the same.
Do you have another suggestions that would work with async code?

Another thing is, the only chunk that's allowed to have its size changed is the last one, the other ones can't have the size changed and there's a validation in the code to avoid it from happening.

ericmackrodt:
Hello, do you have a suggestion for this case?

uriahcarpenter:
@ericmackrodt Ray is on vacation until January 20.

rnicholus:
>What you suggested would be ideal, the problem is that it wouldn't support asynchronous code within onUploadChunk

Yes, it would. As I mentioned in my comment, the internal code would need to change to handle a promissory return value from the callback handler.

ericmackrodt:
Do you prefer that I open a feature request instead of writing the feature?

rnicholus:
If you feel comfortable making an attempt, by all means. But this is likely a pretty complicated change for anyone not familiar with the codebase. Otherwise, just open up a feature request.

Issue #1438, opened 2015-07-22
Content Security Policy compliance


rnicholus:
In order for Fine Uploader to be used in environments that utilize a CSP in an attempt to prevents XSS attacks. Some of this will need to coincide with removal of IE7-related hacks from the codebase, thus this is a set of breaking changes and will need to be part of a planned major release.

A running list of adjustments I'll need to make include:
- [ ] Removal of any `eval` calls (only ever used in IE7 anyway)

Things we probably cannot remove or may require a fair amount of work to remove:
- [ ] Use of data URIs (used to display scaled images)

Items in the second list will probably require documentation. For example, if we don't remove use of data URIs in Fine Uploader, we should note that all web applications that include Fine Uploader must include `img-src 'self' data:;` in their policy.

The documentation piece of this will also probably need to reveal some implementation details. For example, if an integrator has a CSP in place but using Fine Uploader S3, they will need to be sure their CSP allows requests to the S3 bucket's API endpoint.

PR #1528, opened 2016-02-16
Remove eval() fallback for JSON parsing.

KevinGrandon:
Currently eval() is only used for IE7 and prevents sites from implementing content security policies. Since IE7 is a legacy browser, I suggest releasing a fine-uploader version without IE7 support, allowing those who wish to keep legacy support the ability to use an older version of this library.

Refs #1438


rnicholus:
This won't have any affect on CSP, since that code path is only executed in IE7.

Please have a look at the milestone 6.0 case list. You'll see that j have plans to remove a number of legacy code artifacts. There is also an open case in that milestone to address any CSP violations (wherever possible).

Issue #1406, opened 2015-05-05
It's not possible to supply sessionToken if request is being signed by backend


eugenebond:
Hi,

I have the following configuration
```
  <script type="text/javascript">
    $(function() {
      $('#fineUploader').fineUploaderS3({
          template: 'qq-template',
          request: {
              endpoint: "https://MY-BUCKET-HERE.s3.amazonaws.com",
          },
          credentials: {
              accessKey: "...",
              sessionToken: "...",
              expiration: "2015-05-06",
          },
          signature: {
              endpoint: "/my_end_point"
          },
          uploadSuccess: {
              endpoint: "index.html"
          },
          chunking: {
              enabled:   true,
              mandatory: true,
          }
      });
    });
  </script>
```

The headers that end up being POST-ed to the backend are the following
```
POST

image/gif

x-amz-acl:private
x-amz-date:Tue, 05 May 2015 15:11:11 GMT
x-amz-meta-qqfilename:600-animated.gif
/MY-BUCKET-HERE/9c061d9f-c743-4831-93cd-d4fd7cc72896.gif?uploads
```

Since `x-amz-security-token` is missing here, this never gets authenticated by Amazon. I tried plugging `sessionToken` in different places, e.g. into `request` but never managed to get it to work.

Also on a side note, for the example above, the header that gets sent to Amazon is 
`Authorization:AWS null:signture-goes-here`. So it appears that `credentials.accessKey` does not get picked up because it considers `credentials` set to be incomplete without `secretKey` in it (which I don't want to put there, I want the request to be signed by backend). I am not sure if it's a bug or intended behavior. 

I managed to get all of this working in one of two ways:
- by supplying all 4 values into `credentials`
- by having `request.accessKey` set and hacking JS code to set `x-amz-security-token` like this `headers["x-amz-security-token"] = 'a string with my token'`

Version: 5.2.1

rnicholus:
All properties of the `credentials` option, other than `accessKey` are only intended to be used by client-side signing workflow. If you'd like to sign requests on your server, do not supply any `credentials` property other than `accessKey` in your options, and provide a path to your signature server.

eugenebond:
@rnicholus I don't entirely understand how the workflow can work in case of a temporary credentials and if I want to use my own signature server. In this case there is no way to supply `sessionToken` and JS will never add `x-amz-security-token` header which will lead to AWS rejecting the request.

rnicholus:
[...]

PR #1549, opened 2016-03-31
Adding S3 onSigningRequestComplete Event Callback

jfwblog:

I added a new event for which a user can configure an event handler callback,  `onSigningRequestComplete` .  This event gets triggered after fineuploader receives a response from a signing request.  The callback function gets passed 3 parameters, they are `response`, `success` and `xhrOrXdr` (this was modeled after the core event `onSessionRequestComplete`).  The return of the callback will be used to overwrite `response` in case the callback needs to modify the `response` that will be used subsequently by the caller.  If the callback returns a falsey value then an error will be triggered in the caller as if the signing request itself has failed.

I made a minor change to `setCredentials` to allow for certain "required" fields to be omitted if the parameter `ignoreEmpty` is true.  This allows for the `setCredentials` function to be used to set a new `accessKey` without specifying the other fields, which currently `setCredentials` will require that all fields are specified even if they are already undefined or not needed.

These features were added to allow for a signing request to also provide an `accessKey` for the fineuploader to use.  If the `onSigningRequestComplete` event callback makes a call to `setCredentials` and changes the credentials' `accessKey` or `sessionToken` then those new credentials will be used by fineuploader instead for the request to Amazon which is about to occur.

I have tested the changes in Firefox

All unit tests pass according to `grunt test:firefox`.  I believe there is a false positive js indentation warning which I could not resolve.

The pull request is against develop
## Contributor Agreement

**If you are contributing any code whatsoever, you _must_ agree to the terms below. Please indicate your agreement by typing your name below this line.**
jfwblog

In order to clarify the intellectual property license granted with Contributions from any person or entity, Widen must have a Contributor License Agreement (“CLA”) on file that has been signed by each Contributor, indicating agreement to the license terms below. This license is for your protection as a Contributor as well as the protection of Widen; it does not change your rights to use your own Contributions for any other purpose.

You accept and agree to the following terms and conditions for Your present and future Contributions submitted to Widen. Except for the license granted herein to Widen and recipients of software distributed by Widen, You reserve all right, title, and interest in and to Your Contributions.
1. Definitions.
   “You” (or “Your”) shall mean the copyright owner or legal entity authorized by the copyright owner that is making this Agreement with Widen. For legal entities, the entity making a Contribution and all other entities that control, are controlled by, or are under common control with that entity are considered to be a single Contributor. For the purposes of this definition, “control” means (i) the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (ii) ownership of fifty percent (50%) or more of the outstanding shares, or (iii) beneficial ownership of such entity.

“Contribution” shall mean any original work of authorship, including any modifications or additions to an existing work, that is intentionally submitted by You to Widen for inclusion in, or documentation of, any of the products owned or managed by Widen (the “Work”). For the purposes of this definition, “submitted” means any form of electronic, verbal, or written communication sent to Widen or its representatives, including but not limited to communication on electronic mailing lists, source code control systems, and issue tracking systems that are managed by, or on behalf of, Widen for the purpose of discussing and improving the Work, but excluding communication that is conspicuously marked or otherwise designated in writing by You as “Not a Contribution.”
1. Grant of Copyright License. Subject to the terms and conditions of this Agreement, You hereby grant to Widen and to recipients of software distributed by Widen a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare derivative works of, publicly display, publicly perform, sublicense, and distribute Your Contributions and such derivative works.
2. Grant of Patent License. Subject to the terms and conditions of this Agreement, You hereby grant to Widen and to recipients of software distributed by Widen a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable (except as stated in this section) patent license to make, have made, use, offer to sell, sell, import, and otherwise transfer the Work, where such license applies only to those patent claims licensable by You that are necessarily infringed by Your Contribution(s) alone or by combination of Your Contribution(s) with the Work to which such Contribution(s) was submitted. If any entity institutes patent litigation against You or any other entity (including a cross-claim or counterclaim in a lawsuit) alleging that your Contribution, or the Work to which you have contributed, constitutes direct or contributory patent infringement, then any patent licenses granted to that entity under this Agreement for that Contribution or Work shall terminate as of the date such litigation is filed.
3. You represent that you are legally entitled to grant the above license. If your employer(s) has rights to intellectual property that you create that includes your Contributions, you represent that you have received permission to make Contributions on behalf of that employer, that your employer has waived such rights for your Contributions to Widen, or that your employer has executed a separate Corporate CLA with Widen.
4. You represent that each of Your Contributions is Your original creation (see section 7 for submissions on behalf of others). You represent that Your Contribution submissions include complete details of any third-party license or other restriction (including, but not limited to, related patents and trademarks) of which you are personally aware and which are associated with any part of Your Contributions.
5. You are not expected to provide support for Your Contributions, except to the extent You desire to provide support. You may provide support for free, for a fee, or not at all. Unless required by applicable law or agreed to in writing, You provide Your Contributions on an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied, including, without limitation, any warranties or conditions of TITLE, NON- INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A PARTICULAR PURPOSE.
6. Should You wish to submit work that is not Your original creation, You may submit it to Widen separately from any Contribution, identifying the complete details of its source and of any license or other restriction (including, but not limited to, related patents, trademarks, and license agreements) of which you are personally aware, and conspicuously marking the work as “Submitted on behalf of a third-party: [[]named here]”.
7. You agree to notify Widen of any facts or circumstances of which you become aware that would make these representations inaccurate in any respect.

Thank you!

### DIFF
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -265,6 +265,7 @@ qq.s3.RequestSigner = function(o) {
             pendingSignatureData = pendingSignatures[id],
             promise = pendingSignatureData.promise,
             signatureConstructor = pendingSignatureData.signatureConstructor,
+			oldCredentials = Object.assign({}, credentialsProvider.get()),
             errorMessage, response;
 
         delete pendingSignatures[id];
@@ -302,8 +303,6 @@ qq.s3.RequestSigner = function(o) {
             errorMessage = "Received an empty or invalid response from the server!";
         }
 
-        var old_credentials = Object.assign({}, credentialsProvider.get());
-
         response = options.onSigningRequestComplete(response, !isError, xhrOrXdr);
 
         isError = isError || (!response) && Boolean(errorMessage = "onSigningRequestComplete event callback returned false");
@@ -319,10 +318,11 @@ qq.s3.RequestSigner = function(o) {
             generateHeaders(signatureConstructor, response.signature, promise);
         }
         else {
-            if(old_credentials.accessKey != credentialsProvider.get().accessKey || old_credentials.sessionToekn != credentialsProvider.get().sessionToken){
-                options.log('Amazon credentials changed after onSigningRequestComplete event callback');
+            if (oldCredentials.accessKey !== credentialsProvider.get().accessKey || oldCredentials.sessionToekn !== credentialsProvider.get().sessionToken) {
+                options.log("Amazon credentials changed after onSigningRequestComplete event callback");
                 promise.success(response, credentialsProvider.get().accessKey, credentialsProvider.get().sessionToken);
-            }else{
+            }
+			else {
                 promise.success(response);
             }
         }
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -318,7 +318,7 @@ qq.s3.RequestSigner = function(o) {
             generateHeaders(signatureConstructor, response.signature, promise);
         }
         else {
-            if (oldCredentials.accessKey !== credentialsProvider.get().accessKey || oldCredentials.sessionToekn !== credentialsProvider.get().sessionToken) {
+            if (oldCredentials.accessKey !== credentialsProvider.get().accessKey || oldCredentials.sessionToken !== credentialsProvider.get().sessionToken) {
                 options.log("Amazon credentials changed after onSigningRequestComplete event callback");
                 promise.success(response, credentialsProvider.get().accessKey, credentialsProvider.get().sessionToken);
             }
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -62,6 +62,7 @@ qq.s3.RequestSigner = function(o) {
             }
             else {
                 headers.Authorization = "AWS " + credentials.accessKey + ":" + signature;
+                credentials.sessionToken && (headers["x-amz-security-token"] = credentials.sessionToken);
             }
 
             promise.success(headers, signatureConstructor.getEndOfUrl());
@@ -269,7 +270,7 @@ qq.s3.RequestSigner = function(o) {
             signatureConstructor = pendingSignatureData.signatureConstructor,
             oldCredentials = Object.assign({}, credentialsProvider.get()),
             tempCredentials = {},
-            credentials, errorMessage, response;
+            credentials, errorMessage, response, i;
 
         delete pendingSignatures[id];
 
@@ -308,7 +309,22 @@ qq.s3.RequestSigner = function(o) {
 
         response = options.onSigningRequestComplete(response, !isError, xhrOrXdr, tempCredentials);
 
-        if (tempCredentials && Object.keys(tempCredentials).length !== 0) {
+        if (tempCredentials) {
+            if (Object.keys) {
+                Object.keys(tempCredentials).length === 0 && (tempCredentials = null);
+            }
+            else {
+                /** this is for pre-ECMA 5 compatability */
+                for (i in tempCredentials) {
+                    if (tempCredentials.hasOwnProperty(i)) {
+                        tempCredentials = null;
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (tempCredentials) {
             options.log("Temporary Amazon credentials found after onSigningRequestComplete event callback");
             credentials = tempCredentials;
         } else {
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -47,9 +47,11 @@ qq.s3.RequestSigner = function(o) {
         credentialsProvider,
 
         generateHeaders = function(signatureConstructor, signature, promise, credentials) {
-            credentials || (credentials = options.signatureSpec.credentialsProvider.get());
-
             var headers = signatureConstructor.getHeaders();
+
+            if (!credentials) {
+                credentials = options.signatureSpec.credentialsProvider.get();
+            }
 
             if (options.signatureSpec.version === 4) {
                 headers.Authorization = qq.s3.util.V4_ALGORITHM_PARAM_VALUE +
@@ -62,7 +64,10 @@ qq.s3.RequestSigner = function(o) {
             }
             else {
                 headers.Authorization = "AWS " + credentials.accessKey + ":" + signature;
-                credentials.sessionToken && (headers["x-amz-security-token"] = credentials.sessionToken);
+
+                if (credentials.sessionToken) {
+                    headers["x-amz-security-token"] = credentials.sessionToken;
+                }
             }
 
             promise.success(headers, signatureConstructor.getEndOfUrl());
@@ -268,9 +273,15 @@ qq.s3.RequestSigner = function(o) {
             pendingSignatureData = pendingSignatures[id],
             promise = pendingSignatureData.promise,
             signatureConstructor = pendingSignatureData.signatureConstructor,
-            oldCredentials = Object.assign({}, credentialsProvider.get()),
+            credentials = credentialsProvider.get(),
+            oldCredentials = Object.assign ? Object.assign({}, credentials) : {
+                accessKey: credentials.accessKey,
+                secretKey: credentials.secretKey,
+                expiration: credentials.expiration,
+                sessionToken: credentials.sessionToken
+            },
             tempCredentials = {},
-            credentials, errorMessage, response, i;
+            errorMessage, response, i;
 
         delete pendingSignatures[id];
 
@@ -309,9 +320,16 @@ qq.s3.RequestSigner = function(o) {
 
         response = options.onSigningRequestComplete(response, !isError, xhrOrXdr, tempCredentials);
 
+        if (!response) {
+            errorMessage = (errorMessage ? errorMessage + ";  " : "") + "onSigningRequestComplete event callback returned false";
+            isError = true;
+        }
+
         if (tempCredentials) {
             if (Object.keys) {
-                Object.keys(tempCredentials).length === 0 && (tempCredentials = null);
+                if (Object.keys(tempCredentials).length === 0) {
+                    tempCredentials = null;
+                }
             }
             else {
                 /** this is for pre-ECMA 5 compatability */
@@ -322,16 +340,12 @@ qq.s3.RequestSigner = function(o) {
                     }
                 }
             }
-        }
 
-        if (tempCredentials) {
-            options.log("Temporary Amazon credentials found after onSigningRequestComplete event callback");
-            credentials = tempCredentials;
-        } else {
-            credentials = credentialsProvider.get();
+            if (tempCredentials) {
+                options.log("Temporary Amazon credentials found after onSigningRequestComplete event callback");
+                credentials = tempCredentials;
+            }
         }
-
-        isError = isError || (!response) && Boolean(errorMessage = "onSigningRequestComplete event callback returned false");
 
         if (isError) {
             if (errorMessage) {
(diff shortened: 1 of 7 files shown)


jfwblog:
The Travis CI build failure is due to the whitespace issues I mentioned I believed were false positives.  If the `--force` parameter is used to ignore the whitespace issues then the rest of the tests succeed.

rnicholus:
Looks like the whitespace JSCS failures are not false positives. After I look get a chance to look closer at your code and proposal, I'll fix the whitespace issues.

rnicholus:
This will also need some comprehensive unit tests. If you can add those, that will speed up this process. Otherwise, after reviewing, I'll write tests when time allows.

jfwblog:
The whitespace issues have been resolved.  

I can't guarantee that I will have the time to add any new unit tests, I would need to spend considerable time figuring out how to write these tests within the context of the existing test framework.  You or someone else can probably do this more quickly and easily, in more consistent style too.

rnicholus:
Cool, thanks for doing that. Your pull request introduces a lot of changes, which will take me a bit of time to work through. There are a dizzying number of workflows to consider, some of which are sadly not completely covered by unit tests.

Your changes point out a deficiency in Fine Uploader. It's a solid library, but not terribly extensible. And adding more hooks for a particular feature set requires making changes to the entire library. This adds more risk to the changes, and makes it harder for those outside of the project (like yourself) to contribute. I'm very slowly working on the next version of Fine Uploader, which I'm currently calling ["Modern Uploader"](https://github.com/FineUploader/modern-uploader). The goal is to address this issue of extensibility, among other things.

jfwblog:
I made a few more commits that address certain issues and add some additional capabilities.

I added a fourth parameter to the `onSigningRequestComplete` callback which is to be used as a return parameter.  An empty credentials object is passed to the callback which temporary credentials can be assigned to in order to change the credentials used by fineuploader for the current request in progress.  This differs from setting the credentials globally to fineuploader via the `setCredentials` call in that concurrent uploads may all use distinct credentials without risking changing any of the credentials used by any other request (race condition).

In the process of testing this it was also discovered that to support multipart uploads the `generateHeaders` function would need to be updated to accept an additional parameter, `credentials` which influences what credentials are used for the request to Amazon.  This parameter is optional and if omitted then the function will use the `credentialsProvider` like usual.  Also, in the event that `sessionToken` is changed by the callback, `generateHeaders` also required the addition of a `x-amz-security-token` header in order to be accepted by Amazon.  (These details are internal and not necessary to be known by a developer implementing `onSigningRequestComplete` callbacks that change credentials.)

rnicholus:
>I added a fourth parameter to the onSigningRequestComplete callback which is to be used as a return parameter

Why not just check the return value from the callback handler?


I'm going to go through a first pass review now. I'll need to take some time later to look at the changes a bit more closely, write some tests, and take some time manually testing these changes. Documentation will also need to be updated. Please follow this case for progress & comments.

client/js/s3/request-signer.js
```diff
@@ -46,20 +46,23 @@ qq.s3.RequestSigner = function(o) {
         },
         credentialsProvider,
 
-        generateHeaders = function(signatureConstructor, signature, promise) {
+        generateHeaders = function(signatureConstructor, signature, promise, credentials) {
+            credentials || (credentials = options.signatureSpec.credentialsProvider.get());
```
rnicholus: I like how this eliminates some of the boilerplate in this method, but I think the following is a bit more readable:
```javascript
var credentials = credentials || options.signatureSpec.credentialsProvider.get());
```

client/js/s3/request-signer.js
```diff
@@ -46,20 +46,23 @@ qq.s3.RequestSigner = function(o) {
...
                     qq.s3.util.getCredentialsDate(signatureConstructor.getRequestDate()) + "/" +
                     options.signatureSpec.region + "/" +
                     "s3/aws4_request," +
                     "SignedHeaders=" + signatureConstructor.getSignedHeaders() + "," +
                     "Signature=" + signature;
             }
             else {
-                headers.Authorization = "AWS " + options.signatureSpec.credentialsProvider.get().accessKey + ":" + signature;
+                headers.Authorization = "AWS " + credentials.accessKey + ":" + signature;
+                credentials.sessionToken && (headers["x-amz-security-token"] = credentials.sessionToken);
```
rnicholus: This is a bit too dense. I'd reformat as an `if` statement.

client/js/s3/request-signer.js
```diff
@@ -265,7 +268,9 @@ qq.s3.RequestSigner = function(o) {
             pendingSignatureData = pendingSignatures[id],
             promise = pendingSignatureData.promise,
             signatureConstructor = pendingSignatureData.signatureConstructor,
-            errorMessage, response;
+            oldCredentials = Object.assign({}, credentialsProvider.get()),
```
rnicholus: This will have to be removed/changed. `Object.assign` is not supported in any version of IE. Also, Safari, Chrome, and Firefox didn't add support until recently.

client/js/s3/request-signer.js
```diff
@@ -302,6 +307,32 @@ qq.s3.RequestSigner = function(o) {
...
+        }
+
+        if (tempCredentials) {
+            options.log("Temporary Amazon credentials found after onSigningRequestComplete event callback");
+            credentials = tempCredentials;
+        } else {
+            credentials = credentialsProvider.get();
+        }
+
+        isError = isError || (!response) && Boolean(errorMessage = "onSigningRequestComplete event callback returned false");
```
rnicholus: This is pretty dense hard to follow. Also, I'm not sure about determining if an error exists based on text in the error message.

client/js/s3/uploader.basic.js
```diff
@@ -157,7 +158,19 @@
         },
 
         setCredentials: function(credentials, ignoreEmpty) {
-            if (credentials && credentials.secretKey) {
+            if (credentials && ignoreEmpty) {
+                if (this._currentCredentials === undefined) {
+                    this._currentCredentials = {};
+                }
+
+                credentials.accessKey !== undefined && (this._currentCredentials.accessKey = credentials.accessKey);
```
rnicholus: These are a bit too dense and hard to follow, especially the 3rd line. These can probably be expanded into `if` statements and/or moved into other functions.

jfwblog:
> Why not just check the return value from the callback handler?

That would be simple, but I already have the return value being used to return the singing request response.  The response gets fed further down the pipeline into code executed later so I figured it may be useful for the callback to give the user an opportunity to modify the response, in case it has fields which need to be stripped, or transformed, etc. .  So to preserve that functionality I thought passing the optional temporary credentials as an output parameter would be the cleanest way to move forward.  I figured it would be more useful more often to return a modified response than it would be to return credentials.

client/js/s3/request-signer.js
jfwblog: this and other areas have been updated to reflect making the code less dense, easier to follow.

this part here with the `isError` detection was a little convoluted, my apologies.  content of `errorMessage` was not being used to determine error state, rather it was `!response` which if true would ensure that `errorMessage` gets assigned properly (and after being cast to boolean `errorMessage === !response === true`).

But I have made this more clear and better now, nevermind what it was before.

client/js/s3/request-signer.js
jfwblog: I have changed this so now `Object.assign` is used if it is available, if not then object cloning is performed manually with hardcoded fields.

I figured it is good to use `Object.assign` if available because in the future if the credentials object gets more fields added then `Object.assign` will pick them up as opposed to manually copying the object which will require us to add the additional fields in the copying code as well.

rnicholus: You should probably just refer to [the `qq.extend` function in utils](http://docs.fineuploader.com/branch/master/api/qq.html#qq.extend).

jfwblog: This was the ideal solution I was hoping existed, but in my cursory glance at the qq util functions I must not have seen this one.  Will update this later.

rnicholus:
This sounds eerily similar to the feature requested in #1406. That case requires an STS token to be passed to S3 along with each request. This STS token may change/expire at any time, and the associated access key may also change as a result. Thoughts @jfwblog?

rnicholus:
I can see eliminating the 4th param you added and instead checking for a `credentials` property in the response. This would allow the server to include a `credentials` property w/ a new STS token and/or a new access key, each as sub-properties of `credentials`. I don't see the client making these types of changes to the credentials when the request is signed server-side.

I also do not think `setCredentials` should be called at all, to be honest. As you noted, there are potential race conditions that can surface, especially with concurrent chunking enabled. Instead, I would expect Fine Uploader to use the credentials provided during uploader initialization for every request _unless_ the server returns a new set of credentials for a specific request. If a new set of credentials is to be used for the next 10 requests, then the signature server must return these credentials as part of the signature response for each request, otherwise Fine Uploader will use the initial set of credentials. 

Your PR has made me think about this quite a bit, and the need to allow for a constantly changing STS token and access key is something we are looking at for our apps that use Fine Uploader internally as well. I think the issue that will ultimately dictate changes to Fine Uploader is #1406, as this describes an enterprise workflow that is becoming increasingly common as it is seen as a "best practice". That is, associating an EC2 server with an IAM instance profile that contains a set of temporary credentials for signing the request. Fine Uploader could very well be sending requests to a cluster of EC2 instances, each with a different set of temporary credentials. This means each signature request for a single file could require different set credentials for each chunk/request. It seems important for Fine Uploader to support this workflow.

I'm going to keep this PR open, but ultimately I see myself making changes that mirror the specific uses outlined in #1406 once I have taken some time to research this a bit more thoroughly.

jfwblog:
Sorry for the late response:

Essentially, the features requested in #1406 are satisfied with this pull request.  As you have noted, it is not a good idea to use `setCredentials`, my pull request does not call `setCredentials`.  (The confusion is, my pull request accounts for the case where a user does call `setCredentials` from within the callback.  If a user does do this the new credentials should still be used.  If the user instead returns new credentials without calling `setCredentials`, then the callback will use the returned credentials without calling `setCredentials`.  The returned credentials will not persist beyond the current request.)

The scenario described by #1406 is essentially the same scenario which prompted this development.  We were in need of an S3 uploader which uses a different credential for each request.  We wanted the signing server to return new credentials each time it was queried.

The fourth return param for temporary credentials can be eliminated in the way you describe, by incorporating it as a new field on the return object.  That would work, I only went this route because I wanted to avoid conflicts/collisions with the signing request response object.  What if the signing request response object has a field called `credentials` but there is no signing request callback configured?  It should be up to the callback to dictate whether new credentials should be used, the signing server should work with the callback to make the proper arrangements to prevent false positives or other potential issues within fineuploader.

rnicholus:
>What if the signing request response object has a field called credentials but there is no signing request callback configured?

I would expect the response to contain a root `credentials` property for Fine Uploader S3 to use. Fine Uploader S3 would then look for sub-properties matching [the documented `credentials` option](http://docs.fineuploader.com/branch/master/api/options-s3.html#credentials) and apply them automatically to the associated request. I don't see the need for a signing request callback in this instance. Fine Uploader S3 should be able to parse the response and apply the credentials without any integration code.

Thanks for your update. I am currently looking into this as part of a 5.9.0 release. I have a few more pressing adjustments for 5.8.0. And 5.7.0 was just released today.

PR #1585, opened 2016-06-13
Feature/1195 Allow pause of upload queue

Erazihel:

Add 2 methods in the API to pause and unpause the upload queue. These methods will set/unset a flag that is checked during the file upload. If set to true, the upload won't be triggered and the file will be added to the storedFiles.

This feature is required in our current project. We are generating thumbnails by adding photos to the traditional FineUploader queue. The thumbnails are then uploaded by a S3 FineUploader, and the photos using the traditional one. Thing is, we need to upload the thumbnails in priority, and thus pause the traditional upload queue (because of the browser limitation) that will be unpaused once all the thumbnails are sent.

These changes have been tested on Chrome 51.0.2704.84 (64-bit) on OS X El Capitan 10.11.2

Yes

This pull request is against develop


rnicholus:
Thanks for creating this against develop, for your work, and your time.

>Thing is, we need to upload the thumbnails in priority

Can you expand upon this a bit? What determines "priority" here?

>because of the browser limitation

Which limitation are you referring to, and why is this an issue if your app?

It would be helpful if you show your Fine Uploader integration code so it is more obvious how you are using these new API methods.

Erazihel:
A user can upload many photos to our servers that need to be analyzed to sort some informations out of it (which ones are similar or duplicate, which ones can be gathered by theme, etc.).

As soon as the user finish uploading his photos, he should be able to keep going during his experience as fast as possible, and this isn't possible until the thumbnails are generated, uploaded and analyzed.

The photos are not needed for the user to continue his experience (thumbnails are sufficient). They will be, however, needed for the final product that will be created by the user though. That's why we'll upload them once the thumbnails are all sent.

Note that a user can upload thousands of photos. The waiting time will be way to long if we upload all these photos, it will be however acceptable for the thumbnails upload. I think you got the idea why we need to send the thumbnails first.

Also note that the user can upload photos multiple times. It means that if he selects, let's say 50 photos, the thumbnails are generated, uploaded and analyzed, then the photos start to be uploaded. If the user goes back to the photo selection and selects 50 new photos, we have to pause the upload of the photos, treat these 50 new photos, generate, upload and analyze the thumbnails, then unpause the upload of all the photos.

Now, about the browser limitations, I invite you to visit this [post](http://stackoverflow.com/a/985704/4693496). You can only have a certain amount of simultaneous connections to a server, depending on your browser. Average is 5 simultaneous connections/server.

As we need the thumbnails to be sent in priority, we can't afford the upload of the photos to use some of this simultaneous connection, as it will slow down the thumbnails upload. By using the maximum simultaneous connections/server for our thumbnails only, we could reduce the waiting time of the user twice. For thousands of thumbnails, it's not negligible.

Here is the pseudo synchronous code we'd like to use once the user as selected all the photos he wants to upload:

    uploader.addFiles(photosSelected);

    uploader.pauseQueue();

    uploader.generateThumbnails();

    // Use S3 Uploader
    uploaderS3.addFiles(generatedThumbnails);

    // Once the thumbnails are all sent
    uploader.unpauseQueue();

Thank you for your time.

rnicholus:
I'm pretty sure I have a good understanding of your workflow now. You are using a single Fine Uploader instance that handles uploads to S3 and a traditional endpoint (all.fine-uploader.js). Is this correct? _Or_ are you using two instances: 1 S3 and 1 traditional? Let's get that cleared up before I go further.

Erazihel:
We are using two instances, 1 S3 and 1 traditional.

rnicholus:
In that case, I'm not seeing a need for your pull request. 

In the traditional uploader instance, it seems you can simply set the [`autoUpload` option](http://docs.fineuploader.com/branch/master/api/options.html#autoUpload) to `false`. This will allow the queue to build up as files are added, but no files will be uploaded until you call the [`uploadStoredFiles` API method](http://docs.fineuploader.com/branch/master/api/methods.html#uploadStoredFiles).

The S3 uploader can use the default `autoUpload` option, which is `true` - allowing files to be uploaded as they are added (or queued if no connections are free). Once the [`onAllComplete` callback](http://docs.fineuploader.com/branch/master/api/events.html#allComplete) is hit on the S3 uploader, you can then start the traditional uploader by calling `uploadStoredFiles`.

Erazihel:
I thought about your solution, there is still a problem in the case I mentioned before:

> Also note that the user can upload photos multiple times. It means that if he selects, let's say 50 photos, the thumbnails are generated, uploaded and analyzed, then the photos start to be uploaded. If the user goes back to the photo selection and selects 50 new photos, we have to pause the upload of the photos, treat these 50 new photos, generate, upload and analyze the thumbnails, then unpause the upload of all the photos.

We still need the pause feature for this.

rnicholus:
I see. That makes sense, and I now see the need for this. But your proposed solution doesn't solve this problem at all. The internal `_uploadFile` method is called for each file as soon as it is submitted if `autoUpload` is `true`. If `autoUpload` is `false`, as soon as `uploadStoredFiles` is called, then all files are  sent to `_uploadFile`. In other words, if you have submitted 50 files and at least one is currently being uploaded `_uploadFile` has likely already been called for all of the submitted files. 

If you want to prevent any further queued files from being uploaded, you'll need to instead asset some control over the logic in the [`start` method of the upload-handler-controller module](https://github.com/FineUploader/fine-uploader/blob/5.9.0/client/js/upload-handler/upload.handler.controller.js#L558). And to allow the controller to once again attempt to loop through the queued files and attempt to send them, you'll need to call `start` again, passing the ID of the file that you "paused" on. This may be tricky to implement without a good understanding of this module, especially since other in-progress files will attempt to call `start` for the next file once the upload completes. This is why unit tests for these features are important.

Erazihel:
Yes I've tested it manually and it works fine. But you're right, it works because the only reason the queue is paused is that new files have been stored (explaining why the `uploadStoreFiles` has an effect on the paused files). Bringing the modifications ASAP.

Erazihel:
Hi @rnicholus, did you have the time to checkout my updated code?

client/js/uploader.basic.api.js
```diff
@@ -337,6 +343,20 @@
...
+        resumeQueue: function() {
+            this._isPausedQueue = false;
+
+            // If upload has been paused, resume from the id it has been paused
+            // This will also upload all the files in the _waiting queue
+            if (this._pausedId !== -1) {
+                this._uploadFile(this._pausedId);
+            }
+
+            if (this._storedIds.length) {
```
rnicholus: I'm not sure I understand why this action is being taken here. `uploadStoredFiles` applies to an entirely different behavior, and is not really related to pausing an in-progress queue. The decision to upload "stored" files should be made by the integrator through the API.

rnicholus:
I added a comment. Before we can move forward, some unit tests that exercise this new behavior are needed. Any changes to the internal upload queuing code make me nervous since the introduction of concurrent chunking, which made all of the logic in there much more complex.

Erazihel:
All right, fair enough. I just wanted to know if this logic was okay for you before writting unit tests. I'll write them ASAP. Thank you for your feedback.

Erazihel:
Hi @rnicholus, any feedback on these unit tests would be appreciated

rnicholus:
I will take a peek ASAP. Thanks for the reminder.

rnicholus:
Tests are passing and the code looks fine at first glance. Can you update your branch from develop next? Then I'll do some manual testing. If that all checks out, docs are next and then it can be merged into develop.

rnicholus:
Is there anything left to do here? If not, then I'll start manual testing. You can begin with documentation updates if you'd like. Feel free to ask if direction is needed.

Erazihel:
The only thing left to do is documentation so yes, you can start manual testing. I'll take care of the documentation and will come back to you if needed.

rnicholus:
Sorry just a couple more questions for you before I test further:

1. Are you using these exact changes in production yet?
2. Have you had the opportunity to test this and existing related features (chunking, concurrent chunking, etc) in multiple browsers?

Erazihel:
1. Not yet, but we planned a release with these exact changes in production next week
2. We hadn't the opportunity as we're not using any other related features

rnicholus:
Thanks for the update, and apologies for the delayed response. I'll run some manual tests today. Please let me know once you have gone live with this.

rnicholus:
After some manual testing, existing functionality appears to work as it did before, though I still would like to run a few more tests. However, I _am_ seeing some issues with the new pause/resume queue feature. The underlying cause is duplicate entires for a paused file ID in the upload handler controller's `_waiting` array. I can reproduce it fairly easily a couple different ways.

One way to reproduce involves turning on the "concurrent chunking" feature:

1. Ensure `chunking.enabled` is `true`. 
2. Ensure `chunking.concurrent.enabled` is `true`.
3. Drop 3 or more files (dropping more files seems to make the issue easier to reproduce.
4. Pause the queue before the first file completes.
5. After the first file completes, resume the queue.
6. Pause the queue before the second file completes.
7. After the second file completes, resume the queue.

Repeating the last two steps for each file will eventually result in a scenario where the queued files cannot be uploaded (they are "stuck"). In some cases, calling resume multiple times may "fix" the issue, other times not. After setting breakpoints, the `_waiting` array inside upload.handler.controller.js has duplicate entries for the latest queued file.

This can also be reproduced without concurrent chunking:

1. Assuming the `maxConnections` option is set to `3`, drop 6 files.
2. Pause the queue before any of the first 3 files completes.
3. Resume the queue once one of the first 3 files completes.

Repeat the pause/resume steps above until the files become "stuck".

Erazihel:
You're absolutely right, pausing the queue right after resuming it will not reset the paused variables. I moved the unreachable code to the `resumeQueue` function and it works fine without concurrent chunking.

However, I can't test the "concurrent chunking" feature because our servers don't support it. Can you give me any feedback on this ?

client/js/upload-handler/upload.handler.controller.js
```diff
@@ -590,6 +584,7 @@ qq.UploadHandlerController = function(o, namespace) {
     };
 
     qq.extend(this, {
+        connectionManager: connectionManager,
```
rnicholus: I'm not sure I like making `connectionManager` in its entirely available to all code outside of this module. I really don't think anything other than `connectionManager` should touch its internal state.

Erazihel: Yup, my bad. Fixing it right now

client/js/upload-handler/upload.handler.controller.js
```diff
@@ -687,6 +686,10 @@ qq.UploadHandlerController = function(o, namespace) {
             handler.reset();
         },
 
+        resetPausedUpload: function() {
```
rnicholus: much better! I'll re-run manual tests when i get a chance. not sure i will have time today, but we shall see

rnicholus:
Looks like that specific issue is fixed, but I'm seeing another as I test scaled image uploading. With the feature turned on, if you select a bunch of photos, the pause the queue after the first few have upload, a call to resumeQueue doesn't resume the remaining files.

Erazihel:
I manage to reproduce this issue from time to time. It appears that for some reason the `connectionManager` stops being available at some point. Calling the `_uploadFile` method with the `this._pausedId` will only result in adding the file to the `queued` files and thus doesn't resume the remaining ones.

I'm not quite sure how to correct this issue. Any help would be appreciated.

Erazihel:
FYI, we released these changes in production wednesday, so far so good.

rnicholus:
I suspect this will require a non trivial amount of time to track down and fix, which was my concern from the start. I'm not sure when I will be able to take a closer look. Unfortunately, I have a number of other tasks on this project that take priority.

gavboulton:
@Erazihel @rnicholus This feature looks like something I need. I'd like to pause an queued uploads when I receive a 500 response from one upload. I would then like to resume the queue when triggered by the user.

Do you have any plans to further investigate the remaining problem? I could take a look if not. Do you know what the test would be so that I can make it pass?

Erazihel:
@gavboulton I don't have any plans anymore since I've left the company I worked in where this feature was needed.

If you want to take a look at it, go ahead but I can't really help you out though.

rnicholus:
I'm lacking the time to even respond to support/issue requests as of late, so I don't see this progressing with my help anytime soon.

Issue #1602, opened 2016-06-26
Option to set SAS for Azure uploads and skip signing endpoint


birarda:
## Type of issue [REQUIRED]
feature

## Uploader type [REQUIRED]
Azure

## Feature request details

With the S3 version of FineUploader, it is possible to avoid signing requests from your server by using STS and setting those credentials on the FineUploader instance. I'd like to do something similar for Azure.

Could there be a way to a pass a container SAS to the uploader at the beginning of the session and have it use that for all operations in the Azure container directly?

rnicholus:
The workflow you are suggesting would likely be a great deal of work to implement for Azure, as it was for S3. Certainly it would be more than just allowing a SAS to be passed in for all operations. Azure requires, to my knowledge, a newly generated SAS for _each_ request, just as S3 needs a request-specific signature. So, Fine Uploader would have to generate this SAS for each request client-side based on a supplied storage account key (and I assume storage account name would be needed as well). But this sort of operation would be unwise client-side without the ability to generate short-lived heavily locked-down access keys. I personally don't use Azure, so I'm not sure if this type of system even exists.

Providing support for an entirely client-side signing workflow for S3 was challenging, and there is still more to do in terms of STS support for S3 (see #1406 and related cases for details). That said, I don't see anything like this making it onto my personal schedule for development anytime soon. But I would be willing to be available for anyone who is dependent on Azure and knowledgeable in that regard that is interesting in creating a pull request to add alternate signing workflow support to the Azure module.

birarda:
Hey @rnicholus, thank you for the prompt reply. It is much appreciated. You've got a great library here and I hope you don't feel like I was suggesting this would be a trivial amount of work.

I believe it is possible to generate a single SAS that allows all blob operations on the container with read, write, create, list, and delete permissions. I've just tested that now and confirmed that I was able to put an empty block in my Azure container. 

On our end, we're creating that SAS using a stored access policy and then revoking said policy once the user is no longer allowed to upload to their container.

I'm going to attempt to hack this in to a fork of FineUploader today by simply returning a set SAS where it would normally hit the middleman server to request a SAS for the blob.

rnicholus:
>I believe it is possible to generate a single SAS that allows all blob operations on the container with read, write, create, list, and delete permissions

How are you able to prevent the user from using this SAS to delete or list any/all files in the storage account? 

[...]

PR #1604, opened 2016-06-26
Allow a sasQuery to be specified for Azure to avoid the use of a signing endpoint

birarda:

Allows passing of `signature.sasQuery` to Azure options. If no `signature.endpoint` is present this will use the passed query string to immediately create a SAS URI for all requests, without needing the signing endpoint to produce a SAS URI for each request.

This is what I needed to perform what I was after in #1602.

Chrome 52 on OS X

We set a stored access policy on a container and then create a SAS URI for that container with the identifier for the stored access policy and full permissions to the container. That URI, with changes to the path for the blob that actions are being take on, allows all action we need on the given container.

Yes, though I have yet to add tests for the passing of `signature.sasQuery`.

Yes

### DIFF
--- a/client/js/azure/azure.xhr.upload.handler.js
+++ b/client/js/azure/azure.xhr.upload.handler.js
@@ -113,10 +113,17 @@ qq.azure.XhrUploadHandler = function(spec, proxy) {
                 promise.failure({error: "Problem communicating with local server"}, getSasXhr);
             },
             determineBlobUrlSuccess = function(blobUrl) {
-                api.getSasForPutBlobOrBlock.request(getSasId, blobUrl).then(
-                    getSasSuccess,
-                    getSasFailure
-                );
+                // When we already have a SAS query to use then return it here without actually making a request
+                // to the default signature endpoint
+                if (signature.sasQuery && signature.endpoint == null) {
+                    getSasSuccess(blobUrl + "?" + signature.sasQuery);
+                } else {
+                    api.getSasForPutBlobOrBlock.request(getSasId, blobUrl).then(
+                        getSasSuccess,
+                        getSasFailure
+                    );
+                }
+
             },
             determineBlobUrlFailure = function(reason) {
                 log(qq.format("Failed to determine blob name for ID {} - {}", id, reason), "error");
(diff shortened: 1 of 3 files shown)


rnicholus:
Can you explain how you are creating this SAS URI? I'm curious as to how this is limited in terms of scope and time.

rnicholus:
@birarda I'd like to understand a bit how you are creating this SAS URI? This change will need to be documented further on the Azure feature page for other users, so I'd like to be sure I understand the best way to make use of your new API methods without sacrificing security.

rnicholus:
I haven't heard back in a while, so I'm going to close this. But I would be happy to re-open once a discussion regarding best practices in the context of security can be started.

birarda:
I'm sorry about my delay in responding here. I'll try to cover how I build the SAS URI and what we use it for.

In our backend, I generate a SAS URI for a given container. Here are the important params we add to the URI.

`sp (signedpermissions): racwdl` - grants all actions (read, write, add, create, delete, and list) on the container and its contained blobs
`sr (signedresource): c` - specifies that the shared resource is a container
`sv (signedversion): 2015-04-05` - specifies the desired Azure API version
`si (signedidentifier): $stored_access_policy_name` - specifies the access policy to be granted (already stored with the container ACL)
`sig (signature): $signature` - the HMAC-SHA256 signature constructed following https://msdn.microsoft.com/en-us/library/azure/dn140255.aspx

We generate a stored access policy for each container that is valid for one month from creation. This stored access policy is left on the container for as long as we want the SAS URI to be usable. This allows us to revoke access whenever (by removing the stored access policy) even if the client has saved the SAS URI we pass down to their browser.

rnicholus:
What prevents someone from taking the sas URI to delete all items in the container?

birarda:
For our purposes, we want the user that has the SAS URI to be able to delete whatever items they want in the container. Each container is associated with a particular user, so until we revoke their access (by removing the stored access policy) we want them to have that access.

One could opt not to provide that access by removing `d` from the abbreviated permission list.

rnicholus:
Ah, I see. Thanks for the explanation. I'll look over your changes a bit as soon as I can. Can you provide some unit tests? We can use your explanation in this PR as part of the documentation for this feature.

PR #1610, opened 2016-07-08
fix(image): Fixed a problem where image previews weren't being loaded

daviesgeek:
I fixed a problem where the if the thumbnail url returned from the server contained a query string, the image type defaults to jpeg, thus breaking any other file type previews

Chrome on OS X

Yes

Yes


rnicholus:
👏 Thanks for catching and fixing this one!

rnicholus:
I might just end up releasing 5.10.1 with your two fixes next week, just so other have access to these in published releases.

daviesgeek:
No problem.

Awesome! That would be great!

rnicholus:
5.10.1 just released.

PR #1625, opened 2016-07-25
Added "content-length-range" to s3/util.js

davidtgq:
Added `content-length-range` parameter to `UNPREFIXED_PARAM_NAMES` in `s3/util.js`

The problem that this solves:
Using `validation.sizeLimit` changes `content-length-range`, but also prevents files greater than that size to be added to FineUploader at all. This prevents FineUploader from accepting a large image and compressing it until it is within the size limit. Setting `content-length-range` in `request.params` avoids this problem.

Related:
http://stackoverflow.com/questions/38569985/how-to-set-content-length-range-in-fineuploader-s3

None

I don't yet know how to run tests unfortunately :(

No


rnicholus:
We can and should talk about this more - I think a few changes are needed. But before we do that, can you close this PR and open a new one against the develop branch?

PR #1626, opened 2016-07-25
Added "content-length-range" to s3/util.js

davidtgq:

Added `content-length-range` parameter to `UNPREFIXED_PARAM_NAMES` in `s3/util.js`

The problem that this solves:
Using `validation.sizeLimit` changes `content-length-range`, but also prevents files greater than that size to be added to FineUploader at all. This prevents FineUploader from accepting a large image and compressing it until it is within the size limit. Setting `content-length-range` in `request.params` avoids this problem.

Related:
http://stackoverflow.com/questions/38569985/how-to-set-content-length-range-in-fineuploader-s3

None

I don't yet know how to run tests unfortunately :(

yes


rnicholus:
Looks like tests are passing, but I'd like to see a couple more changes before this can be merged into develop. Unfortunately, nothing is ever as simple as it appears in a project of this size and complexity.

1. Upper-case for each segment in the added header: `Content-Length-Range`. 
2. Manual tests using this "header" for chunked and non-chunked uploads.

The 2nd item is needed as `Content-Length-Range` is a bit strange as it's only a policy document conditional and not a supported or recognized HTTP header. Without further changes to the code, I suspect that the bucket's CORS policy will need to be updated to allow for this request header, even though `Content-Length-Range` means nothing in the context of an S3 API call. I am currently not sure if we should just leave this as-is and expect any integrator who sets this param/header to know that the CORS policy must be updated, or if we should prevent this from being sent as an HTTP header.

Issue #1016, opened 2013-10-16
5 - Support uploads to S3 via CloudFront distribution


rnicholus:
The current plan is to support simple (non-chunked) uploads to a CloudFront distribution.  Chunked uploads are currently not possible when targeting a CloudFront distribution since CloudFront rips off the Authorization header containing the signature before forwarding the request on to S3.  The Authorization header is a required field in the request when using any of the S3 multipart upload REST calls, which are needed to support Fine Uploader S3's chunking and auto-resume features.  I have opened a [request in the CloudFront forums](https://forums.aws.amazon.com/thread.jspa?threadID=137626) asking for this behavior to be modified so multipart uploads requests can target a CloudFront distribution.

The planned support for this is still mostly undetermined, and it is also not determined if this support will be part of 4.0.  Making this part of 4.0 is possible, but looking less likely as I run into issues with CloudFront's handling of upload-related requests.  I'm currently struggling to get path patterns for upload requests to work.  I've opened another thread in the forum detailing my issue at https://forums.aws.amazon.com/thread.jspa?threadID=137627&tstart=0.

cdanzig:
@rnicholus in the process of exploring chunked uploading to S3 via CloudFront. Were you ever able to get the pieces of the puzzle together?

rnicholus:
Uploads to CloudFront _should_ work, but there appear to be some issues with CF ripping some headers off the REST requests that prevented me from certifying support for CF.  Amazon never acknowledged the issue though, even though I am able to upload files to S3 via a 3rd-party CDN without issue.

rbliss:
@rnicholus Talked to the CloudFront guys directly. They're now saying that ALL the headers should pass through appropriately. Can you still recreate the issue or have a curl command to recreate the issue?

rnicholus:
They told me the same thing a little while back. It isn't true, as far as I can tell. I opened up [a support request in December, detailing the issue](https://forums.aws.amazon.com/thread.jspa?threadID=167234). The sum it up, sending upload requests to S3 using the multipart upload API via CloudFront still results in removed headers. For example, x-amz-acl and all x-amz-meta headers are removed. Sending the exact same request directly to S3, or via a different CDN (such as fastly) did not result in any problems. The issue was specifically with CloudFront.

cmwright:
It doesn't help anyone trying to use CloudFront, but we were able to get uploads via CDN working with Edgecast if someone is looking for a Fastly alternative.

[...]

PR #1627, opened 2016-07-26
docs(features): Add S3 Transfer Acceleration to S3 feature

### LINKED ISSUE #1556: Support S3 Transfer Acceleration
This _should_ just work, simply by setting the endpoint option to "BUCKET_NAME.s3-accelerate.amazonaws.com" and turning on the feature at the bucket level, but AWS is notorious for oversimplifying their announcements and leaving large gaping holes their new offerings. So this will require testing and an update in docs. The new S3 feature is documented (initially) at https://aws.amazon.com/blogs/aws/aws-storage-update-amazon-s3-transfer-acceleration-larger-snowballs-in-more-regions.

sohkai:

Adds documentation changes to indicate support for S3 Transfer Acceleration (see #1556).

N/A; docs change.

N/A; docs change.

No; docs change.

---

Fixes #1556. May also handle #1016 (just use the Transfer Accelerate option for now), although maybe in the last year Cloudfront has fixed their issues.

See additional comments based on changes.

We've been having success with this option on version 5.3, so I'd assume, based on the changes, that it should work for any version from at least 5.1 onwards. Maybe @DaveDeCaprio, @e-tip, or @ludofleury could chime in with which versions they've been using?

### DIFF
--- a/docs/features/s3.jmd
+++ b/docs/features/s3.jmd
@@ -10,33 +10,31 @@
 Please see [the blog post on Fine Uploader S3][s3blog].
 
 
-### Uploading to S3 With Transfer Acceleration
-Amazon's [S3 Transfer Acceleration option][s3transferacceleration] is by far the easiest way to have your
-uploads to S3 go through a CDN (Amazon Cloudfront). To start using this feature, simply enable it in your AWS
-console and use `https://[bucket-name].s3-accelerate.amazonaws.com` instead of `https://[bucket-name].s3.amazonaws.com`
-in the [`request.endpoint` option][requestendpoint].
-
-
-### Uploading to S3 Through a CDN Without Transfer Acceleration
-Support for uploads to an S3 bucket via a CDN was added in Fine Uploader 5.1.0.  You can upload files to
+### Uploading to S3 Through a CDN
+Support for uploads to an S3 bucket via a CDN was added in Fine Uploader 5.1.0. You can upload files to
 any S3 bucket through any CDN, provided the CDN forwards all headers and does not append any additional
 headers to the request that is sent on to S3. An example of a CDN that should work fine is [fastly][fastly].
-Ironically, an example of a CDN that does not appear to work is Amazon Cloudfront (see [issue #1016 in the
-Fine Uploader GitHub issue tracker][issue1016]). If you'd like to use Amazon Cloudfront as a CDN for your
-uploads, please use the [S3 Transfer Acceleration option][s3transferacceleration] instead.
 
-#### Overview
-When uploading to S3 via a CDN, you must specify the name of the bucket.  This is not required for uploads
-directly to S3, since Fine Uploader is able to determine the bucket name by examining the S3 endpoint URL.
-This is obviously not the case for a CDN endpoint.  So, in addition to specifying the CDN endpoint via the
-[`request.endpoint` option][requestendpoint], you must also specify a bucket name via the [`objectProperties.bucket` option][objectpropertiesbucket].
+#### S3 Transfer Acceleration
+Amazon's [S3 Transfer Acceleration option][s3transferacceleration] is by far the easiest way to have your
+uploads to S3 go through a CDN (Amazon Cloudfront). To use this feature, simply enable it in your AWS console
+and specify `https://[bucket-name].s3-accelerate.amazonaws.com` instead of `https://[bucket-name].s3.amazonaws.com`
+for the [`request.endpoint` option][requestendpoint].
+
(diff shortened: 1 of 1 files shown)


docs/features/s3.jmd
```diff
@@ -10,21 +10,20 @@
...
-POSTs to an S3 bucket appear to fail when the endpoint is a non-default CloudFront behavior.  Third, there
-does not appear to be any way to forward all headers to an S3 origin via a CloudFront endpoint.  This is
-a problem for uploads using S3's REST API, since we must include the ACL, metadata, encryption, redundancy
-parameters, and a number of other information via the Initiate Multipart Upload POST request.  We have notified
-Amazon of these issues, but they have been unwilling to address or acknowledge them.  For continuing updates
-on support to CloudFront, please see [issue #1016 in the Fine Uploader GitHub issue tracker][issue1016].
+headers to the request that is sent on to S3. An example of a CDN that should work fine is [fastly][fastly].
+Ironically, an example of a CDN that does not appear to work is Amazon Cloudfront (see [issue #1016 in the
+Fine Uploader GitHub issue tracker][issue1016]). If you'd like to use Amazon Cloudfront as a CDN for your
+uploads, please use the [S3 Transfer Acceleration option][s3transferacceleration] instead.
```
SohKai: Wasn't sure how to link to the previous heading with jmd syntax so I linked it to the AWS page for now. If possible, I think it'd be nicer if this link pointed to the section above.

docs/features/s3.jmd
```diff
@@ -10,21 +10,20 @@
...
 
 #### Overview
 When uploading to S3 via a CDN, you must specify the name of the bucket.  This is not required for uploads
```
SohKai: Didn't check if `When uploading to S3 via a CDN, you must specify the name of the bucket` applies to the accelerated endpoint since it's using a different subdomain. Maybe @rnicholus you could quickly double check how this logic is applied right now and if it'll work for the `s3-accelerate.amazonaws.com` subdomain?

docs/features/s3.jmd
rnicholus: Which specific section are you trying to link to?

docs/features/s3.jmd
rnicholus: You mentioned that you were able to use Fine Uploader S3 w/ transfer acceleration. Did you have to specify an `objectProperties.bucket` option?

rnicholus:
I think the best approach here is to completely remove the existing CloudFront-related documentation, and include a new sub-section under the main/existing "Uploading to S3 Through a CDN" section that talks about S3 Transfer Acceleration. Once #1016 has been verified, then we can add CF docs as part of closing that case.

docs/features/s3.jmd
rnicholus: I would expect that an explicit `objectProperties.bucket` option is _not_ required here. The use S3 Transfer Acceleration, the URL must always be bucketName.s3-accelerate.amazonaws.com, and this is not a 3rd-party CDN. That said, I'm still curious if you had to specify this option yourself.

ludofleury:
I was using the latest version of FineUploader.

About #1016, I really think that this AWS accelerate feature remove the needs to do fancy thing with CloudFront "PUT". (On a side note, #1016 is actually working when configured without OAI).

But really, except if you upload more than 350 TB per month, I don't see a reason to not use accelerate instead of CloudFront.

rnicholus:
Thanks for your input @ludofleury. Perhaps the best course of action here is to:

1. Close #1016 in favor of S3 Transfer Acceleration support.
2. Remove mention of CloudFront in the docs, as I mentioned in my previous comment.

ludofleury:
Yes, I totally agree on this, #1016 becomes obsolete (as the feature implemented).
I will try soon the accelerate feature with signed request (we have this requirement) and if I have any issue, I will get back there.

docs/features/s3.jmd
SohKai: We specify the `objectProperties.bucket` out of habit, but can confirm that it still works when not specified. I'll add a bit here to clarify that you also don't need to specify the bucket if using Accelerate.

docs/features/s3.jmd
SohKai: The `Uploading to S3 With Transfer Acceleration` section, but this point will be moot with the changes below.

SohKai:
I've updated with the suggestions above, will squash after everything checks out.

rnicholus:
Looks pretty good to me. No need to worry about squashing. I can squash at merge time.

PR #1665, opened 2016-09-20
Don't requeue uploads in the waiting queue.

fromkeith:

Stop the connection manager from pushing an existing id into the waiting queue.

If the user calls retry twice (or any other weird race condition) the connection manager re-queues the file. Then when 'free' is called, we see that this id exists in the waiting queue. Thus we don't free the open slot or invoke the next item for upload.
This can cause the uploader to stop working entirely as the open slots are never cleared.

Chrome on Windows 10.

Yes.

Yes. Going to Develop. Did discover this issue when using 5.10.

Fixing a bug where if 'retry' is invoked on an item already in the waiting queue it gets re-queued. Which means that after the upload finishes and 'free' is called, we incorrectly don't upload the next item.

### DIFF
--- a/client/js/upload-handler/upload.handler.controller.js
+++ b/client/js/upload-handler/upload.handler.controller.js
@@ -322,7 +322,11 @@ qq.UploadHandlerController = function(o, namespace) {
 
         open: function(id, chunkIdx) {
             if (chunkIdx == null) {
-                connectionManager._waiting.push(id);
+                var alreadyInWaiting = qq.indexOf(connectionManager._waiting, id);
+                var alreadyInOpen = qq.indexOf(connectionManager._open, id);
+                if (alreadyInWaiting === -1 && alreadyInOpen === -1) {
+                    connectionManager._waiting.push(id);
+                }
             }
 
             if (connectionManager.available()) {


rnicholus:
Thanks so much for detecting this issue and contributing a fix. I'll run some manual regression tests to confirm that everything is still working as expected (since upload.handler.controller.js is both critical and complex). If everything checks out, I'll merge this in.

Also, can you please sign the CLA?

rnicholus:
Can you please merge develop back into this branch?

rnicholus:
I'm going to push this to 5.11.10, as I'd like to get 5.11.9 out as soon as possible and I have yet to even start on some of those bug fixes.

rnicholus:
This is still on my radar, but 5.11.9 is going to be my focus first. This fix is scheduled for 5.11.10.

rnicholus:
Looks like a _lot_ of changes have been added to this branch, unrelated to the original problem. Was this by mistake?

fromkeith:
wow... yea...
thats not supposed to be in there..

fromkeith:
Fixed hopefully...

rnicholus:
Yep, thanks. Trying to be as transparent as possible: the number of cases for me to look into/verify is piling up, and, unfortunately, the amount of time I have to dedicate to this project is rapidly decreasing.

rnicholus:
Going to have to push this to 5.11.11. I have to hotfixes ready for 5.11.10, and I'd like to release that today. I haven't gotten around to testing this one yet, and it's more risky than the others. At least one unit test needs to be written for this anyway, and I'm not sure when I'll get to that, so if you are able to contribute one, that will speed things up a bit.

Issue #743, opened 2013-03-14
Timeout for bad internet connection


Maigre:
Hi there !
Thanks for your uploader, it's a really big help :)

The problem: 
I try to upload big files over bad internet connection (chunks are usefull !)
but if my connection drop in the middle of the upload, 
the ajax POST of the current chunk doesn't recieve answer from the server,
and it continue to wait... (regardless if the connection is back or not)
is it possible to set up a timeout on the ajax waiting, 
so if the connection has dropped it start to retry (in case the connection is back)
and finally push an error if it is still impossible to reach the server 
so the client can check his connection and then manually retry ?

Thanks !

Best Regards,

Thomas

rnicholus:
XHR timeouts are now supported in Chrome/Safari, so perhaps we should look into implementing this.

rnicholus:
I'm going to look into this as part of 5.11.9, since network errors (particularly pulling the connection mid-upload) causes the upload to hang for a _long_ time. That seems like a bug to me, and if we can make this less of an issue, even in a subset of browsers, that seems like a big win.

rnicholus:
I think I now remember why I kicked this can down the road for so long. Setting a timeout on an `XMLHttpRequest` instance does not function as one would expect (or at least as I would expect). The clock starts ticking as soon as the request is sent. If the request does not complete before the timeout expires, the browser cancels the request. I would expect the request to timeout only if there is no HTTP traffic for a length of time equal to the timeout value. 

The problem with the standardized behavior of `XMLHttpRequest.timeout` is that it may be very tricky to determine an appropriate and effective value. When you choose a timeout value, you have to consider that some internet connections are faster than others. For example, if you choose a timeout value of 5 seconds, with a chunk size of 5 MB, the request will _always_ timeout for files being sent from a typical home internet connection with available upstream bandwidth of 4 Mbps, since each 5 MB chunk will take about 10 seconds to complete.

In light of this, I'm strongly considering abandonment of this "feature". Another option: advise integrators to set a reasonably high value, but this honestly seems like it may cause more problems than it solves. I'm open to input, but I'm putting this on hold for now.

PR #1666, opened 2016-09-22
Reject the signing promises on failure.

fromkeith:
If signing a file for s3 upload fails FineUploader hangs. Promises related to signing were not calling failure on their parent promises.

Repo steps:
1. Find a large file.
2. Put it in an SD card.
3. Add the file to FineUploader
4. Pull out SD card before the signing can finish.
5. Signing methods fail, rejecting their promise.
6. Promise rejection is not propagated up.

Windows 10. Chrome

Yes

Yes. Develop.

### DIFF
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -402,6 +402,8 @@ qq.s3.RequestSigner = function(o) {
                 requestInfo.hashedContent = hashedContent;
 
                 promise.success(generateStringToSign(requestInfo));
+            }, function (err) {
+                promise.failure(err);
             });
         }
         else {
@@ -424,6 +426,8 @@ qq.s3.RequestSigner = function(o) {
 
             toBeSigned.signatureConstructor.getToSign(id).then(function(signatureArtifacts) {
                 signApiRequest(toBeSigned.signatureConstructor, signatureArtifacts.stringToSign, signatureEffort);
+            }, function (err) {
+                signatureEffort.failure(err);
             });
         }
         // Form upload (w/ policy document)
@@ -514,6 +518,9 @@ qq.s3.RequestSigner = function(o) {
                             .withParams(params)
                             .withQueryParams(queryParams)
                             .send();
+                    }, function (err) {
+                        options.log("Failed to construct signature. ", "error");
+                        signatureEffort.failure("Failed to construct signature.");
                     });
                 }
                 else {
@@ -600,6 +607,8 @@ qq.s3.RequestSigner = function(o) {
                             stringToSign: artifacts.toSign,
                             stringToSignRaw: artifacts.toSignRaw
                         });
+                    }, function (err) {
+                        promise.failure(err);
                     });
 
                     return promise;


rnicholus:
Thanks for the PR. I'm a bit surprised that this is broken. I seem to remember this working at some point. I can see how this would fail if the file suddenly disappeared during the upload process.

I'll take a closer look soon.

fromkeith:
I know if the file is pulled before signing, it will have a size of 0 and I think the signing will complete succesfully. It just signs what it thinks an empty file. Then it uploads to s3, and of course s3 complains about an empty file. The root of that problem is that the file size went to zero when it wasn't zero to begin with. I couldn't figure out where to add code to solve that, so instead thought relying on S3s errors would be fine.
The error that I encountered, was the FileReader getting an error when creating the signature. It would properly reject its promise. However, the calls to getEncodedHashedPayload did not listen for the rejection. So I just went in and propagated up the rejection call.

rnicholus:
Interesting - thanks for the details. You are correct about the empty file when it is removed from the filesystem during the upload process. I'll take a closer look when I get a chance.

rnicholus:
I expect to be able to look at this closer tomorrow. If all looks well, I'll merge into develop as part of 5.11.9.

rnicholus:
Can you describe the _symptoms_ of this issue? I'm not exactly sure what I should be seeing when attempting to reproduce.

fromkeith:
I saw the file just 'hang'... It never actually starts uploading or reports an error. No XHR requests get made and the state of the file doesn't change.

rnicholus:
I'm not able to follow the exact steps in your description, but I did notice a couple related issues:

1. When the network connection is pulled, the upload hangs for quite a while. I think this can be solved with an XMLHttpRequest timeout.
2. If the file is deleted while the upload is in progress, fine uploader will continue to send chunks, but they are 0 sized. It seems like Fine Uploader should fail the upload in this case.

I think both of these issues should be addressed as part of this case. I'll look into making this adjustments soon. I hope to release those along with your fix to compete 5.11.9.

fromkeith:
If you put a break point on the 'failure' below, do you ever hit it?

```javascript
getEncodedHashedPayload: function(body) {
    ...
        reader = new FileReader();
        reader.onloadend = function(e) {
            if (e.target.readyState === FileReader.DONE) {
                if (e.target.error) {
                    promise.failure(e.target.error); // here
                }
                else {
                    var wordArray = qq.CryptoJS.lib.WordArray.create(e.target.result);
                    promise.success(qq.CryptoJS.SHA256(wordArray).toString());
                }
            }
        };
   ...
```

I was able to hit the failure if I pulled the sd card while it was trying to read the file.

From what you saw:

1. is something that is especially problematic for really bad internet connections. I would love to see a timeout setting somewhere.
2. Definitely ran into this. Less useless xhr requests would be awesome

rnicholus:
I was never able to hit the breakpoint above. I simply saw FU send a bunch of empty chunks at S3 after the file was deleted. But I believe your changes are correct, so I will merge this in.

I'm going to prioritize #743 and #1669 to address the other two issues we discussed. If you'd like to help out with either, let me know.

Issue #1688, opened 2016-11-09
Usage in TypeScript - Angular 2 based projects


SinghSukhdeep:
## Type of issue
feature request

## Uploader type
All

## Feature request details
Just started using this library and it is a very solid one with many great features.
I'm trying to integrate with TypeScript in Angular 2 project.
I've read the [documentation ](http://docs.fineuploader.com/branch/master/features/modules.html) about how to import as ES6 modules.

the typescript compiler complaints if I use  `import qq from 'fine-uploader'` because there isn't any index file in the root of the library which would export anything as a typical Typescript settings.

if I do `let qq = require('fine-uploader')`, things work fine, but I need to use `import `rather than `require `to be able to generate rollup bundle later.

### Detailed description of the feature
As Angular is very popular Front end framework, it will be great if we're able to use FineUploader with ease.
I can understand that you have less time for features like these.
I'm very new to this library as I just started playing around with it and I have less knowledge about the core functions and the file structure.
What needs to be done here is to add TypeScript definition files and have named exports for all the modules.
You have already done something very similar for **react-fine-uploader**
I'm not sure how much work will it require for Angular 2 - TypeScript Compatibility.

Let me know your thoughts on it and I'll be happy to help If I can.

rnicholus:
Related to #1652?

Sukh9212:
Yes, it's related to #1652 
In addition to having typescript definitions, also needed is class declarations and named exports from common index.d.ts file.

A good example to that is [angular2-highcharts](https://github.com/gevgeny/angular2-highcharts) library, which is just a handful of typescript wrappers for the original commonJS Highcharts library and works perfectly with Angular 2. If you look at their `dist` directory, all it contains is just handful or file which have class declarations and named exports for all the different components.

I know this probably is a low priority task for you. I can help to get this done, but I have very little knowledge about fine-uploader library

rnicholus:
I'm afraid I'm not the best person to implement this feature for a few reasons:

1. I'm not familiar with Angular 2 (I used Angular 1 in previous projects, but have since moved to React).
2. I'm not familiar enough with Typescript, though it's on my list of tech to investigate.
3. As you mentioned, there are a number of other cases higher up on my list.

That said, I do see how this would be a very valuable update to Fine Uploader. I'm willing to provide advice and answer questions at any time if you are willing to participate in coding this up.

[...]

PR #1689, opened 2016-11-14
Adding typescript definition file

SinghSukhdeep:
Adding TypeScript definitions allows integration with TypeScript based projects

N/A

I haven't written tests for this yet

yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,8 @@ cryptojs-files = \
 	$(js-3rdparty-src-dir)/crypto-js/lib-typedarrays.js
 
 jquery-files = \
-	$(js-src-dir)/jquery-plugin.js
+	$(js-src-dir)/jquery-plugin.js \
+	$(js-src-dir)/jquery-dnd.js
 
 dnd-files-only = \
 	$(js-src-dir)/dnd.js
@@ -38,10 +39,6 @@ dnd-files = \
 	$(js-src-dir)/features.js \
 	$(js-src-dir)/promise.js \
 	$(js-src-dir)/dnd.js
-
-dnd-jquery-files = \
-	$(jquery-files) \
-	$(dnd-files)
 
 core-files = \
 	$(js-src-dir)/util.js \
@@ -65,7 +62,6 @@ core-files = \
 	$(js-src-dir)/image-support/megapix-image.js \
 	$(js-src-dir)/image-support/image.js \
 	$(js-src-dir)/image-support/exif.js \
-	$(js-src-dir)/identify.js \
 	$(js-src-dir)/identify.js \
 	$(js-src-dir)/image-support/validation.image.js \
 	$(js-src-dir)/session.js \
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean _build publish
+.PHONY: clean _build publish start-test-resources-server test-resources-server.PID start-root-server root-server.PID
 
 version=$(shell node -pe "require('./package.json').version")
 dist-out-dir = _dist
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![npm](https://img.shields.io/npm/v/fine-uploader.svg)](https://www.npmjs.com/package/fine-uploader)
 [![license](https://img.shields.io/badge/license-MIT-brightgreen.svg)](LICENSE)
 [![Stackoverflow](https://img.shields.io/badge/ask-on%20stack%20overflow-brightgreen.svg)](http://stackoverflow.com/questions/tagged/fine-uploader)
+[![Twitter URL](https://img.shields.io/twitter/url/https/twitter.com/fineuploader.svg?style=social&label=Follow%20%40FineUploader)](https://twitter.com/fineuploader)
 
 [**Documentation**](http://docs.fineuploader.com) |
 [**Examples**](http://fineuploader.com/demos) |
--- a/client/js/upload-handler/upload.handler.controller.js
+++ b/client/js/upload-handler/upload.handler.controller.js
@@ -86,6 +86,61 @@ qq.UploadHandlerController = function(o, namespace) {
             );
         },
 
+        handleFailure: function(chunkIdx, id, response, xhr) {
+            var name = options.getName(id);
+
+            log("Chunked upload request failed for " + id + ", chunk " + chunkIdx);
+
+            handler.clearCachedChunk(id, chunkIdx);
+
+            var responseToReport = upload.normalizeResponse(response, false),
+                inProgressIdx;
+
+            if (responseToReport.reset) {
+                chunked.reset(id);
+            }
+            else {
+                inProgressIdx = qq.indexOf(handler._getFileState(id).chunking.inProgress, chunkIdx);
+                if (inProgressIdx >= 0) {
+                    handler._getFileState(id).chunking.inProgress.splice(inProgressIdx, 1);
+                    handler._getFileState(id).chunking.remaining.unshift(chunkIdx);
+                }
+            }
+
+            // We may have aborted all other in-progress chunks for this file due to a failure.
+            // If so, ignore the failures associated with those aborts.
+            if (!handler._getFileState(id).temp.ignoreFailure) {
+                // If this chunk has failed, we want to ignore all other failures of currently in-progress
+                // chunks since they will be explicitly aborted
+                if (concurrentChunkingPossible) {
+                    handler._getFileState(id).temp.ignoreFailure = true;
+
+                    log(qq.format("Going to attempt to abort these chunks: {}. These are currently in-progress: {}.", JSON.stringify(Object.keys(handler._getXhrs(id))), JSON.stringify(handler._getFileState(id).chunking.inProgress)));
+                    qq.each(handler._getXhrs(id), function(ckid, ckXhr) {
+                        log(qq.format("Attempting to abort file {}.{}. XHR readyState {}. ", id, ckid, ckXhr.readyState));
+                        ckXhr.abort();
+                        // Flag the transport, in case we are waiting for some other async operation
+                        // to complete before attempting to upload the chunk
+                        ckXhr._cancelled = true;
+                    });
+
+                    // We must indicate that all aborted chunks are no longer in progress
+                    handler.moveInProgressToRemaining(id);
+
+                    // Free up any connections used by these chunks, but don't allow any
+                    // other files to take up the connections (until we have exhausted all auto-retries)
+                    connectionManager.free(id, true);
+                }
+
+                if (!options.onAutoRetry(id, name, responseToReport, xhr)) {
+                    // If one chunk fails, abort all of the others to avoid odd race conditions that occur
+                    // if a chunk succeeds immediately after one fails before we have determined if the upload
+                    // is a failure or not.
+                    upload.cleanup(id, responseToReport, xhr);
+                }
+            }
+        },
+
         hasMoreParts: function(id) {
             return !!handler._getFileState(id).chunking.remaining.length;
         },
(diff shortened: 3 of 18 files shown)


rnicholus:
I'll make this as "in progress" for now. Please do let me know when you have a question or need a code review.

rnicholus:
Related #1688.

SinghSukhdeep:
The `element `and `template `options are not listed in core api options, but they are needed if the id of the element is different than default.
I think they should be listed on the api core options page along with others.

Also, in fine-uploader.spec.ts fle, If I use the 'new' keyword to instantiate FineUploader, TypeScript is complaining.
If I don't use the 'new' keyword, things seem to be okay. I'll do more testing once I'm done writing everything

SinghSukhdeep:
http://docs.fineuploader.com/branch/master/api/options-ui.html#thumbnails.placeholders
the `notAvailablePath` option is listed twice

rnicholus:
Thanks for finding that issue in the docs. I'll push a fix for that next. 

Regarding `new` - you definitely have to `new` up a `qq.FineUploader` instance. Not doing so will result in a `function` return type, and, as a result, you won't be able to access any API methods. Also, the internal code expects `qq.FineUploader` to be properly constructed, otherwise the value of `this` will be `window` instead of the `qq.FineUploader` instance.

Not sure if Typescript provides some magic that makes `new` unnecessary. I'm not very familiar with the language myself.

SinghSukhdeep:
http://docs.fineuploader.com/branch/master/api/options-azure.html#signature.endpoint
In description here where it says "The endpoint that Fine Uploader can use to send GET for a SAS before sending requests off to S3"
I think instead of S3 it should be Azure.

rnicholus:
Thanks for catching that. Fixed!

rnicholus:
Note that your branch should be based off of Fine Uploader develop

SinghSukhdeep:
I see that you have changed the base branch from master to develop.
I should have forked from develop.
Is there anything else I need to do to base off of develop branch?
In this thread I'm seeing this message
`Add more commits by pushing to the master branch on SinghSukhdeep/fine-uploader.`
Which branch should I push my changes now?

rnicholus:
it probably doesn't matter much since, at the time you forked, master and develop were identical.

SinghSukhdeep:
http://docs.fineuploader.com/branch/master/api/qq.html#qq.each
the description for parameter `iterable` is incorrect.

rnicholus:
I'd suggest leaving any of the `qq` util functions out. I'd like to remove those entirely in a future major release.

SinghSukhdeep:
Okay, well so far I've written about 80% of util functions already. I'll take them out then.
If there is anything else you have in mind which is to be removed in future releases, please let me know before I write it all.

rnicholus:
Sorry. If you're almost done with util functions, you are free to proceed. I can just remove them once 6.0 is released, which may not be for a while anyway.

I'll fix the docs for `qq.each` too.

rnicholus:
docs are updated

SinghSukhdeep:
I've written down everything that is there in the API's.
Please do a code review when you get a chance.
I'm gonna do some testing meanwhile and will push if I find any issues.
Thanks 😃

rnicholus:
I'll try to take a close look in the next few days. Other than creating the definition files, what else, if anything, needs to be done to ensure Fine Uploader can be used w/ TS in an Angular2 project? I'm also wondering how these TS def files are picked up.

SinghSukhdeep:
Creating a definition file enables the use of FineUploader in any TypeScript project.
Angular 2 is itself written in TypeScript, so this facilitates the use of FineUploader in Angular 2 project as well.

But still, something like `import qq from 'fine-uploader'` won't be possible, because that requires class declarations and named exports. (To do that, I'll need to do some more research).

So now with TS def file, If someone wants to use FineUploader, they can just have the FineUploader JS globally scoped and just start writing TypeScript.

[Here ](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)is how we can publish our TS def file.
Let me know and we'll proceed further.

TS def files are picked up automatically by TypeScript aware editor such as Visual Studio Code (which I've used to write declarations), we just have to publish it properly.

SinghSukhdeep:
TS does support CommonJS modules. 
I can do `let qq = require('fine-uploader')` and it will work fine.
But to use the import syntax, A library needs to have an index file at the root which would do named export of ES6 modules.
I can get an Angular 2 project working using the `require` syntax, but I won't be able to generate a production bundle using Rollup, which needs the `import `syntax rather than `require` to effectively apply tree-shaking.

Having said that, I wouldn't call myself a GURU on those things, but that's what I've figured since I've been working with TS and Angular 2

rnicholus:
>But to use the import syntax, A library needs to have an index file at the root which would do named export of ES6 modules.

The package.json file already handles this.

SinghSukhdeep:
I'm currently trying to follow the TS [module resolution](http://www.typescriptlang.org/docs/handbook/module-resolution.html)
Following that, if I place my TS def file index.d.ts in the `node_module/fine-uploader/` directory 
and place a default export in index.d.ts like this 

``` 
declare var qq: FineUploader.qq;

export default qq;
```
then the `import qq from 'fine-uploader'` seems to be okay.
I need to do some more work/research/testing on it to make sure things go well all the way up to generating a production bundle using Rollup.
I'll keep you posted

rnicholus:
I'm guessing exports will be needed for all Fine Uploader builds, similar to what already exists in the lib directory.

SinghSukhdeep:
Ok so, I've done some basic testing in a dummy Angular 2 app and everything seems to be working fine.
Not doing any named imports/exports as of now (need to do more research on it to make it work), just including fine-uploader.js globally.

rnicholus:
Sorry, haven't gotten around to reviewing this yet. Hopefully over the U.S. holidays I'll find some time to do so.

rnicholus:
something appears to be wrong here. the changed file list seems to be missing a number of files.

SinghSukhdeep:
Okay, I'll push all my changes to the develop branch and change the fork's base branch to develop.
Is that what you're saying?

rnicholus:
Yes, I _think_ that will fix the issue

rnicholus:
doesn't look like that worked - you can see in the "changes" tab at the top of this PR. Looks like a new PR is needed instead.

SinghSukhdeep:
The changed file list shows two files changed (which is all what I've added), I'm not sure what is missing?

rnicholus:
Oh, ok. The def file was not automatically expanded in the diff viewer due to its size, which is probably why I overlooked it.

rnicholus:
Just started to look at this again. Sorry for the delay, but I'm only able to pick up a few minutes here and there for the foreseeable future.  

What exactly is `typescript/fine-uploader.spec.ts`?

rnicholus:
Just started to look at the definition file. I can safely say that I'll never have enough time to review all of that, nor will I be able to verify since I don't use TS in any projects. So I'm included to just merge this through. But I'm wondering - how do projects pick up this definition file? How does TS locate this? Where does it have to reside inside of the package retrieved from npm?

SinghSukhdeep:
The `typescript/fine-uploader.spec.ts` is just a test file showing the use of proper syntax in TS.
About how will TS pickup the def file please read [this](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)
Either we put the def file in a directory and point `"types"` property in `package.json` to that path or we just put the file at the root and name it `index.d.ts` or we publish it separately from the project under Github's `DefinitelyTyped` and then users can install the def file as `npm install @types/fine-uploader`.

I'm currently using it as if installed using third option.
The def file will be then located at `node_modules/@types/fine-uploader/index.d.ts`.
They do recommend the first approach though.

rnicholus:
Fine with the first approach. To complete this:

- [ ] update to package.json as specified in your last message
- [ ] update to Makefile to ensure the proper TS files are included in the zip files & the npm package. It's probably best to include these in the /client/commonJs directory in the source tree. The contents of that dir will be included in the lib/ dir in the npm package. 

I think it would be nice if a feature page was created or something somewhere in the docs was updated to reflect TS support.

Related: Are you interested in becoming a contributor for this project? Seems you now have a better understanding of the API than most at this point.

SinghSukhdeep:
As of TS v2.1, the first approach seems to be working properly (Wasn't working before, that's why I was using @types approach).
I'll review the def file once more and will let you know when I think everything is okay.

And yes I have no issues with becoming a contributor. What do I have to do in order to be so?

SinghSukhdeep:
My bad, I can get it working only by using @types approach.
TS compiler is complaining with other two ways, It has something to do with the way fine-uploader file structure is.
Works only if I put the def file at `node_modules/@types/fine-uploader/index.d.ts`.
I have to do some more digging on this

rnicholus:
@SinghSukhdeep Was wondering if you have any updates on your progress, or if you need anything from me?

SinghSukhdeep:
@rnicholus sorry didn't get any time during the holidays and had some tight deadlines before that, will have some time this week to look at it.

rnicholus:
No rush at all. Just checking in.

noah79:
Please merge this!

rnicholus:
@SinghSukhdeep Just wanted to bring to your attention DefinitelyTyped/DefinitelyTyped#13768. It just came onto my radar a few days ago. It's not clear to me how this affects your efforts here. Was wondering if you had some comments.

SinghSukhdeep:
I looked at it, they are also preparing a similar definitions file as we are.
Maybe somehow we can merge their progress with ours and get someone to review our code.

About removing a package, once we publish fine-uploader with the definitions included, then they can deprecate that package, we don't have to do it.

In this pull I need to set the `types` property to point to our bundled declaration file in order to locate the definition file, which I haven't done yet as I'm unable to get it to work that way.
I've looked at some of the TS docs but haven't found a resolution to the problem yet. I'll keep looking

andy-ms:
Hi @SinghSukhdeep, I'd be willing to help, but could you explain to me the way this library is used? On the DefinitelyTyped tests, everything uses a `qq` global variable. But the NPM package appears to have `module.exports`. Can this library be used as a module?

rnicholus:
The library can be used as a module. Examples/details at http://docs.fineuploader.com/branch/master/features/modules.html.

rnicholus:
After looking at this closer, could it be that the definition file is not included with the packaged library, and that is the cause of this issue? The entire source tree is not published to npm. There is a build step that only includes the bundled code and a subset of the project root. This is something I mentioned in [a comment a while back](https://github.com/FineUploader/fine-uploader/pull/1689#issuecomment-266555976).

SinghSukhdeep:
@andy-ms okay got it, I also came to this resolution because I'm not using the library as a module and I was missing that triple slash directive.
Including the dependency like that does the job.
Thanks for your help.

Any chance someone can review our code before we publish the package?
I would really appreciate that.

bradfordwagner:
@SinghSukhdeep I think you did some really nice thorough work here. 

@rnicholus I think we need to set up CI for this TypeScript component. Treating the TypeScript definitions as first class is much better than having it maintained in the DefinitelyTyped repository, and I'd be happy to contribute. Would it be possible to have a feature branch that both @SinghSukhdeep and I could work out of?

rnicholus:
I'd be happy to add you both as contributors to fine-uploader. Then, this PR could be merged into a local feature branch where work could continue. How does that sound?

bradfordwagner:
Sounds great to me

rnicholus:
I've sent both of you invitations. In order to bring this into the repo without merging it into develop (since it's not complete yet), I think the next steps should be:

1. Create a new branch in this repo based off of develop.
2. Change the target branch of this PR to the branch created in step 1 & merge it in.
3. Open up a new PR and begin to collaborate and make changes on the branch created in step 1. Please reference this PR in the new one.

SinghSukhdeep:
Everything done. New PR #1719

PR #1691, opened 2016-11-20
chore(Makefile): remove duplicate line

guzart:


rnicholus:
Nice catch. This means the `qq.Identify` module is defined twice in the built code. Even though this is a small module and size will only decrease a small amount, I think that merits a hotfix release. Thanks!

Going forward, changes should be initially based off of develop. This is why I have an issue template, which you deleted. But I do thank you again for your help here.

guzart:
Apologize for not following guidelines, I just saw that duplicate line while I was reading the code and went to submit a pull request using Github's UI. Will submit changes from develop next time 😸

rnicholus:
This is going to go out today (likely) as part of 5.11.10. Thanks for your help!

PR #1702, opened 2016-12-16
S3 Signature on Worker

fromkeith:
Have the option for the S3 Signing process (v4) to be done in a worker thread. My browser no longer freezes on large files.

Windows 10
* Chrome 56.0
* Firefox 50.1.0
* Edge 38
* Internet Explorer 11

OSX
* Safari 10.0.2

Yes.
I haven't build the docs though...

Develop

## More Info
You might consider this a scary change :)

A couple things I want to point out, slash get your feedback on.

* The worker can be implemented as either Inline or as a separate file.
* From what I can find IE is the only one that doesn't support the inline worker. 
   * So I've added in a graceful fallback to just the old in 'main thread' method.
* Some people might not want the extra overhead of the inline file.
    * The extra overhead in this case is really the CryptoJS library.
   * I've made the option 'useWorker' to be able to specify a string. That way people can override with their own external worker.
* The inline worker file is generated at compile time. See the __build-s3-inline-worker_ build task.
     * It's done by concatenating 3 files.
         * worker.start.js, CyrptoJS files, worker.end.js
     * A simple script (workerToInline) then converts that file into a string and adds a function to qq.
     * We can now include the inline worker as a build file, as per normal js files.
* A new build artefact is added for the external worker.
* I've added a new S3 output that includes the inline worker. The regular output for s3 should not include it.
* Guidance on how you want the build artefacts to look like would be great. 

Feedback, changes, clarity, etc.. are all welcome.

### DIFF
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -33,7 +33,8 @@ qq.s3.RequestSigner = function(o) {
                 credentialsProvider: {},
                 endpoint: null,
                 customHeaders: {},
-                version: 2
+                version: 2,
+                useWorker: false
             },
             maxConnections: 3,
             endpointStore: {},
@@ -153,10 +154,51 @@ qq.s3.RequestSigner = function(o) {
 
             getEncodedHashedPayload: function(body) {
                 var promise = new qq.Promise(),
-                    reader;
+                    reader,
+                    self = this;
 
                 if (qq.isBlob(body)) {
-                    // TODO hash blob in webworker if this becomes a notable perf issue
+                    // Only if a worker is requested will we load it.
+                    // We will fallback to the inline reader if the worker
+                    // fails to load.
+                    if (options.signatureSpec.useWorker) {
+                        if (!this._worker) {
+                            try {
+                                // allow the worker path to be overwritten.. Eg. we aren't using an inline worker.
+                                var workerUrl = options.signatureSpec.useWorker;
+                                if (workerUrl === true) {
+                                    if (!qq.s3.worker) {
+                                        throw new Error("Missing inline s3 worker.");
+                                    }
+                                    workerUrl = qq.s3.worker();
+                                }
+                                this._worker = new Worker(workerUrl);
+                                this._workerPromises = {};
+                                this._worker.onmessage = function (e) {
+                                    if (!self._workerPromises[e.data.id]) {
+                                        options.log("Worker returned a result for an request we dont know about.");
+                                        return;
+                                    }
+                                    if (e.data.err) {
+                                        self._workerPromises[e.data.id].failure(e.data.err);
+                                    } else {
+                                        self._workerPromises[e.data.id].success(e.data.resp);
+                                    }
+                                    delete self._workerPromises[e.data.id];
+                                };
+                            } catch (ex) {
+                                // worker is not supported or invalid
+                                options.log("Worker failed to be created. Defaulting back to main thread processing.", ex);
+                                options.signatureSpec.useWorker = false;
+                            }
+                        }
+                        if (this._worker) {
+                            var task = {file: body, id: qq.getUniqueId()};
+                            this._workerPromises[task.id] = promise;
+                            this._worker.postMessage(task);
+                            return promise;
+                        }
+                    }
                     reader = new FileReader();
                     reader.onloadend = function(e) {
                         if (e.target.readyState === FileReader.DONE) {
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -269,7 +269,8 @@ qq.s3.RequestSigner = function(o) {
     credentialsProvider = options.signatureSpec.credentialsProvider;
     if (options.signatureSpec.workerUrl !== null) {
         workerManager = new qq.s3.RequestSignerWorkerManager({
-            workerUrl: options.signatureSpec.workerUrl
+            workerUrl: options.signatureSpec.workerUrl,
+            log: options.log,
         });
     }
 
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -33,8 +33,7 @@ qq.s3.RequestSigner = function(o) {
                 credentialsProvider: {},
                 endpoint: null,
                 customHeaders: {},
-                version: 2,
-                workerUrl: null
+                version: 2
             },
             maxConnections: 3,
             endpointStore: {},
@@ -46,7 +45,6 @@ qq.s3.RequestSigner = function(o) {
             log: function(str, level) {}
         },
         credentialsProvider,
-        workerManager,
 
         generateHeaders = function(signatureConstructor, signature, promise) {
             var headers = signatureConstructor.getHeaders();
@@ -159,8 +157,8 @@ qq.s3.RequestSigner = function(o) {
                 if (qq.isBlob(body)) {
                     // We will fallback to the inline reader if the worker was
                     // not loaded correctly
-                    if (workerManager) {
-                        promise = workerManager.generateSignature(body);
+                    if (options.signatureSpec.workerManager) {
+                        promise = options.signatureSpec.workerManager.generateSignature(body);
                         if (promise !== null) {
                             return promise;
                         }
@@ -267,12 +265,6 @@ qq.s3.RequestSigner = function(o) {
 
     qq.extend(options, o, true);
     credentialsProvider = options.signatureSpec.credentialsProvider;
-    if (options.signatureSpec.workerUrl !== null) {
-        workerManager = new qq.s3.RequestSignerWorkerManager({
-            workerUrl: options.signatureSpec.workerUrl,
-            log: options.log,
-        });
-    }
 
     function handleSignatureReceived(id, xhrOrXdr, isError) {
         var responseJson = xhrOrXdr.responseText,
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -34,7 +34,7 @@ qq.s3.RequestSigner = function(o) {
                 endpoint: null,
                 customHeaders: {},
                 version: 2,
-                useWorker: false
+                workerUrl: null
             },
             maxConnections: 3,
             endpointStore: {},
@@ -46,6 +46,7 @@ qq.s3.RequestSigner = function(o) {
             log: function(str, level) {}
         },
         credentialsProvider,
+        workerManager,
 
         generateHeaders = function(signatureConstructor, signature, promise) {
             var headers = signatureConstructor.getHeaders();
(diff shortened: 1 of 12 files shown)


fromkeith:
Looks like lint failures... fixing that

client/js/s3/request-signer.js
```diff
@@ -153,10 +154,51 @@ qq.s3.RequestSigner = function(o) {
...
+                    reader,
+                    self = this;
 
                 if (qq.isBlob(body)) {
-                    // TODO hash blob in webworker if this becomes a notable perf issue
+                    // Only if a worker is requested will we load it.
+                    // We will fallback to the inline reader if the worker
+                    // fails to load.
+                    if (options.signatureSpec.useWorker) {
+                        if (!this._worker) {
```
rnicholus: This file is already quite large. Ideally, this logic would be split out into another JS file.

client/js/s3/request-signer.js
```diff
@@ -153,10 +154,51 @@ qq.s3.RequestSigner = function(o) {
...
+                            try {
+                                // allow the worker path to be overwritten.. Eg. we aren't using an inline worker.
+                                var workerUrl = options.signatureSpec.useWorker;
+                                if (workerUrl === true) {
+                                    if (!qq.s3.worker) {
+                                        throw new Error("Missing inline s3 worker.");
+                                    }
+                                    workerUrl = qq.s3.worker();
```
rnicholus: Not really fond of the name of this function. It's not obvious that a URL is to be returned here, or what the worker does.

client/js/s3/request-signer.js
```diff
@@ -153,10 +154,51 @@ qq.s3.RequestSigner = function(o) {
...
+                    // We will fallback to the inline reader if the worker
+                    // fails to load.
+                    if (options.signatureSpec.useWorker) {
+                        if (!this._worker) {
+                            try {
+                                // allow the worker path to be overwritten.. Eg. we aren't using an inline worker.
+                                var workerUrl = options.signatureSpec.useWorker;
+                                if (workerUrl === true) {
+                                    if (!qq.s3.worker) {
+                                        throw new Error("Missing inline s3 worker.");
```
rnicholus: I think a `qq.Error` is probably more appropriate.

rnicholus (changes requested):
Interesting, and it looks like you put a lot of work into this. I appreciate the effort.

My concern is having yet another JS build artifact. How many bytes does this add to the existing S3 JS build files? Unless the increase is notable, I'd rather just bundle this in with the existing core and UI S3 builds.

Also, I really think unit tests are prudent here, plus some documentation should probably be added to the S3 feature page. I'm happy to assist any way I can, but I'm finding I have less and less time to evolve this library these days.

fromkeith:
I never actually looked at the file size differences...

s3-core.min: 142KB
s3-core-worker.min: 150KB

I'll take a look at unit tests and incorporating your code comments a bit later.

rnicholus:
8K difference - after gzipping, that's almost a wash. Using web workers for this task seems like a big win to me, so I'm leaning towards including it in all existing S3 build files.

fromkeith:
From your feedback:

- 'useWorker' changed to 'workerUrl'
- 'qq.s3.worker()' changed to 'qq.s3.createS3InlineWorkerUrl()'
- Moved the new code from request-signer into a new file 'request-signer.worker-manager.js'
- Added in unit tests. I had to fix Date so that it was constant, so I can verify the signatures not changing.
- Removed the error from being thrown. Instead it is logged.

fromkeith:
I also want to do a bit more testing on multiple browsers with these changes. I've just done a quick one on Firefox, IE, Edge and Chrome.. not as indepth as I want yet. But I thought you would want to look at the reformatting sooner rather than later.

One thing that I could use guidance on... is that because RequestSigner gets created multiple times, my worker-manager gets created multiple times.. Is there a better place to create the manager, but still get the options from the signer?

client/js/s3/request-signer.js
```diff
@@ -153,52 +154,18 @@ qq.s3.RequestSigner = function(o) {
...
-                        }
-                        if (this._worker) {
-                            var task = {file: body, id: qq.getUniqueId()};
-                            this._workerPromises[task.id] = promise;
-                            this._worker.postMessage(task);
+                    // We will fallback to the inline reader if the worker was
+                    // not loaded correctly
+                    if (workerManager) {
+                        promise = workerManager.generateSignature(body);
+                        if (promise !== null) {
```
rnicholus: Is it possible that `promise` can be `undefined` here? If so, this should be `if (promise != null {`.

client/js/s3/request-signer.worker-manager.js
```diff
@@ -0,0 +1,88 @@
+/* globals qq, CryptoJS */
+
+/**
+ * Manages creation and communication of s3 signature workers.
+ */
+qq.s3.RequestSignerWorkerManager = function (o) {
+    "use strict";
+    var _worker = null,
+        _workerPromises = {},
```
rnicholus: curious why this are prefixed with `_`

client/js/s3/request-signer.worker-manager.js
```diff
@@ -0,0 +1,88 @@
...
+    qq.extend(options, o, true);
+
+    function init() {
+        var workerUrl;
+        switch (typeof options.workerUrl) {
+            case "string":
+                if (options.workerUrl !== "inline") {
+                    workerUrl = options.workerUrl;
+                } else {
+                    if (!qq.s3.createS3InlineWorkerUrl) {
```
rnicholus: Why would this ever be falsy?

client/js/s3/request-signer.worker-manager.js
```diff
@@ -0,0 +1,88 @@
...
+                        qq.Error("Missing inline s3 worker");
```
rnicholus: This should be thrown. It's a no-op as it stands.

client/js/s3/request-signer.worker-manager.js
```diff
@@ -0,0 +1,88 @@
...
+                var outstandingRequests = Object.keys(_workerPromises),
+                    i;
+                for (i = 0; i < outstandingRequests.length; i++) {
+                    _workerPromises[outstandingRequests[i]].failure(e);
+                    delete _workerPromises[outstandingRequests[i]];
+                }
+            };
+            _worker.onmessage = function (e) {
+                if (!_workerPromises[e.data.id]) {
+                    options.log("Worker returned a result for an request we dont know about.");
```
rnicholus: seems like this should be a warning or error

client/js/s3/request-signer.worker-manager.js
```diff
@@ -0,0 +1,88 @@
...
+                    return;
+                }
+                if (e.data.err) {
```
rnicholus: Would prefer this to be an `else if` and the empty `return` removed in the above conditional.

docs/api/options-s3.jmd
```diff
@@ -99,7 +99,7 @@ alert("The [`request.customHeaders` option](options.html#request.customHeaders)
...
         ("signature.endpoint", "endpoint", "The endpoint that Fine Uploader can use to send policy documents (HTML form uploads) or other strings to sign (REST requests) before sending requests off to S3.", "String", "null",),
         ("signature.version", "version", "The AWS/S3 signature version to use. Currently supported values are `2` and `4`. Directly related to [`objectProperties.region`](#objectProperties.region).", "Integer", "2",),
-        ("signature.useWorker", "useWorker", "Make client side signature processing for v4 signatures occur in a background worker thread. Set to true to use the inline worker. Otherwise set the URL to the worker to use.", "Boolean/String", "false",),
+        ("signature.workerUrl", "workerUrl", "Make client side signature processing for v4 signatures occur in a background worker thread. Set to 'inline' to use the inline worker. Otherwise set the URL for the worker to use.", "String", "null",),
```
rnicholus: What does the default value of `null` mean?

package.json
```diff
@@ -57,6 +57,7 @@
     "mocha": "3.2.0",
     "node-static": "0.7.9",
     "pica": "latest",
+    "timemachine": "^0.2.8",
```
rnicholus: Can you lock this to a specific version?

workerToInline.js
```diff
@@ -23,7 +23,7 @@ fs.readFile(source, 'UTF-8', function (err, data) {
     let sourceCode = jsStringEscape(data);
 
     let outCode = `
```
rnicholus: I'm very concerned about using template strings and `let`/`const`, since not all bowsers that support web workers support these ES6 language features.

rnicholus (changes requested):
Looking good, but a few changes are needed.

client/js/s3/request-signer.worker-manager.js
fromkeith: Old habits?
I can remove the `_`s

client/js/s3/request-signer.worker-manager.js
fromkeith: Left over from when s3 core didn't include the inline worker. It should be safe to remove it.

client/js/s3/request-signer.worker-manager.js
fromkeith: if we remove the check, that line goes away.

client/js/s3/request-signer.worker-manager.js
fromkeith: I do find short-circuit returning to be easier to read. However, its your code base, so I can modify it to an else-if if you prefer.

docs/api/options-s3.jmd
fromkeith: Off. No worker used.
I've set it as the default so that its by default backwards compatible.

workerToInline.js
fromkeith: That's run by node in the build process. Not by a browser.

rnicholus:
>because RequestSigner gets created multiple times, my worker-manager gets created multiple times.. Is there a better place to create the manager, but still get the options from the signer?

Instead of passing the `workerUrl` as an option to `RequestSigner`, why not construct the `RequestSignerWorkerManager` as part of the S3 upload instance initialization or in `qq.s3.XhrUploadHandler`, and pass the `RequestSignerWorkerManager` instance to `RequestSigner` when it is constructed as part of the `signatureSpec`?

client/js/s3/request-signer.worker-manager.js
rnicholus: Minor detail, but that would be great. Elsewhere in the project, underscores are used to mark publicly-accessible methods/props that are only meant to be used internally (a poor-man's private).

client/js/s3/request-signer.worker-manager.js
rnicholus: Great, please do.

client/js/s3/request-signer.worker-manager.js
rnicholus: Generally speaking, I'm in favor of Dijkstra's approach - single exit. I'm not married to this though. This is just a suggestion.

docs/api/options-s3.jmd
rnicholus: Ok. Please include this info in the option description, if you could.

workerToInline.js
rnicholus: Right, my bad. As long as we stick with Node 5+, we should be good.

rnicholus:
Interested in pushing this through. Are you able to address the remaining issues?

fromkeith:
Sorry. Got distracted by other things.
I think I have a working copy that I've restructured mostly. I think the last main task I had left was moving the worker creation to a different module per your suggestion.

fromkeith:
Anything I need to do to progress?

rnicholus:
Hey there. Sorry for the silence. Reviewing this is on my to-do list, but I have a few things ahead in the queue, and I have some other new commitments that I have to attend to as well.

fromkeith:
No worries. Just wanted to make sure I wasn't blocking anything.

Issue #903, opened 2013-06-28
1 - Allow upload with empty file.


AidasK:
"AutoRun.inf is empty, please select files again without it."
Empty file is also a file and this message bothers users a lot.
It is more for drag and drop uploads, because all uploads get canceled. 
But it would be nice to have empty file upload feature.

To implement this we only need to make optional validation in: 
_validateFileOrBlobData
and change this:

``` javascript
else if (size === 0){
        this._itemError('emptyError', name);
        valid = false;
}
```

to

``` javascript
else if (!this._options.validation.allowEmpty && size === 0){
        this._itemError('emptyError', name);
        valid = false;
}
```

achung-miovision:
Are there any updates to getting this PR merged and released?

rnicholus:
@achung-miovision This isn't a PR

achung-miovision:
My apologies, there's a linked PR https://github.com/FineUploader/fine-uploader/pull/1673 that's been sitting around.

rnicholus:
Remaining in that PR:

- unit tests
- documentation

If someone can tackle those items, I can merge and release after some manual tests on my end.

adrianchung:
@rnicholus I've enhanced the PR and opened up a new PR here: https://github.com/FineUploader/fine-uploader/pull/1710

PR #1710, opened 2017-01-05
Allow upload with empty file

adrianchung:
Implement issue #903 . Allow uploading of 0 byte files. Enhancement on pull request https://github.com/FineUploader/fine-uploader/pull/1673

Chrome 55.0.2883.95 (64-bit) on OSX 10.11.6

yes

yes

### DIFF
--- a/client/js/uploader.api.js
+++ b/client/js/uploader.api.js
@@ -326,7 +326,7 @@
 
             this._templating.updateProgress(id, loaded, total);
 
-            if (Math.round(loaded / total * 100) === 100) {
+            if (total === 0 || Math.round(loaded / total * 100) === 100) {
                 this._templating.hideCancel(id);
                 this._templating.hidePause(id);
                 this._templating.hideProgress(id);
--- a/test/unit/uploader.basic.api.js
+++ b/test/unit/uploader.basic.api.js
@@ -516,21 +516,25 @@ describe("uploader.basic.api.js", function () {
         });
 
         it("fails if file is empty and allowEmpty is false", function() {
+            var result = 1;
             qq.isFileOrInput = function() { return true; };
-            fineuploader._fileOrBlobRejected = function() {};
+            fineuploader._fileOrBlobRejected = function() { result = -1; };
             var validationDescriptor = { size: 0 };
 
             fineuploader._validateFileOrBlobData({}, validationDescriptor)
-                .then(function() { assert.fail(); }, function() {});
+                .done(function() { assert.equal(-1, result); });
         });
 
         it("passes if file is empty and allowEmpty is true", function() {
             fineuploader._options.validation.allowEmpty = true;
+
+            var result = 1;
             qq.isFileOrInput = function() { return true; };
+            fineuploader._fileOrBlobRejected = function() { result = -1; };
             var validationDescriptor = { size: 0 };
 
             fineuploader._validateFileOrBlobData({}, validationDescriptor)
-                .then(function() {}, function() { assert.fail(); });
+                .done(function() { assert.equal(1, result); });
         });
     });
 });
--- a/test/unit/uploader.basic.api.js
+++ b/test/unit/uploader.basic.api.js
@@ -479,4 +479,30 @@ describe("uploader.basic.api.js", function () {
             uploader._handleNewFile(fileInput, 0, []);
         });
     });
+
+    describe("_formatSize", function() {
+        beforeEach(function () {
+            fineuploader = new qq.FineUploaderBasic();
+        });
+
+        it("formats 0 bytes properly", function() {
+            var formattedSize = fineuploader._formatSize(0);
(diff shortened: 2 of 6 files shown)


client/js/uploader.api.js
```diff
@@ -326,7 +326,7 @@
 
             this._templating.updateProgress(id, loaded, total);
 
-            if (Math.round(loaded / total * 100) === 100) {
+            if (total === 0 || Math.round(loaded / total * 100) === 100) {
```
rnicholus: Why was this added? It's not clear to me what consequences are associated with this change.

rnicholus (changes requested):
Looks good, other than my one question. Also, could you write some unit tests that exercise `allowEmpty: true`?

client/js/uploader.api.js
achung-miovision: When files are uploading you see a progress bar, but if there's no bytes to upload then we're technically done, so we need to hide the rest of the icons

achung-miovision:
Unit tests have been added. I'm not sure what the CLA is for as I'm split between two github accounts right now and I've signed it on both. If it's a problem I can dig further to see what's happening there.

test/unit/uploader.basic.api.js
```diff
@@ -505,4 +505,32 @@ describe("uploader.basic.api.js", function () {
...
+            qq.isFileOrInput = originalFileOrInput;
+        });
+
+        it("fails if file is empty and allowEmpty is false", function() {
+            qq.isFileOrInput = function() { return true; };
+            fineuploader._fileOrBlobRejected = function() {};
+            var validationDescriptor = { size: 0 };
+
+            fineuploader._validateFileOrBlobData({}, validationDescriptor)
+                .then(function() { assert.fail(); }, function() {});
```
rnicholus: This (and the subsequent test) do not look valid. As far as I can tell, the test will exit before the promise is resolved. This is effectively an async test. You'll need to make use of the `done` function in both tests.

In the first test, in the second function (failure), add a call to `done`. In the second test, add a call to `done` in the first `then` function (success).

test/unit/uploader.basic.api.js
```diff
@@ -516,21 +516,25 @@ describe("uploader.basic.api.js", function () {
...
         it("fails if file is empty and allowEmpty is false", function() {
+            var result = 1;
             qq.isFileOrInput = function() { return true; };
-            fineuploader._fileOrBlobRejected = function() {};
+            fineuploader._fileOrBlobRejected = function() { result = -1; };
             var validationDescriptor = { size: 0 };
 
             fineuploader._validateFileOrBlobData({}, validationDescriptor)
-                .then(function() { assert.fail(); }, function() {});
+                .done(function() { assert.equal(-1, result); });
```
rnicholus: I would suggest reverting these changes and making use of the `done` function instead. You can use this test as an example: https://github.com/adrianchung/fine-uploader/blob/40db9e3652ea73cf13d7d5244a3336994afb88a5/test/unit/uploader.basic.api.js#L195

test/unit/uploader.basic.api.js
```diff
@@ -479,4 +479,62 @@ describe("uploader.basic.api.js", function () {
...
+        });
+
+        it("fails if file is empty and allowEmpty is false", function() {
+            var result = 1;
+            qq.isFileOrInput = function() { return true; };
+            fineuploader._fileOrBlobRejected = function() { result = -1; };
+            var validationDescriptor = { size: 0 };
+
+            fineuploader._validateFileOrBlobData({}, validationDescriptor)
+                .done(function() { assert.equal(-1, result); });
```
adrianchung: I decided to leverage the failed callback here to set the result. fine uploader seems to have its own promise implementation, so I can't actually call done(); from without the success/failure callbacks. I flipped the values and verified that the tests pass as expected.

test/unit/uploader.basic.api.js
achung-miovision: Oh I see, I was confused between promise done() and mocha async done for when the test completes.. updated as requested.

test/unit/uploader.basic.api.js
rnicholus: Not sure I follow. You can (and should) call `done` inside of either the failure or the success `then` callbacks. The tests are likely passing as the tests are exiting prematurely. Promises are asynchronous operations, and your tests are written as if they were synchronous.

adrianchung: Please ignore this comment. I understand what you mean now that I've looked up your example and your mocha reference properly

rnicholus (approved):

rnicholus:
I'll just run a few quick manual tests sometime in the next few days or so. If all looks good, I'll release this as 5.12.0. Thanks for your work on this!

rnicholus:
This has been merged into develop. I hope to release v5.12.0 this weekend. Thanks again!

PR #1714, opened 2017-01-10
handle 204 no content situation

Darmody:
Give default value `'{}'` to json parser for 204 response

All browsers

yes

Sorry, not sure what this question mean? Maybe yes.

### DIFF
--- a/client/js/traditional/traditional.xhr.upload.handler.js
+++ b/client/js/traditional/traditional.xhr.upload.handler.js
@@ -100,7 +100,7 @@ qq.traditional.XhrUploadHandler = function(spec, proxy) {
 
             try {
                 log(qq.format("Received response status {} with body: {}", xhr.status, xhr.responseText));
-                response = qq.parseJson(xhr.responseText);
+                response = qq.parseJson(xhr.responseText || "{ \"success\": \"true\" }");
             }
             catch (error) {
                 upload && log("Error when attempting to parse xhr response text (" + error.message + ")", "error");


rnicholus:
I'm fairly certain that this change is not sufficient, and an empty response will still be marked as a failed request. See the `isErrorUploadResponse` function.

Darmody:
@rnicholus Hi.

https://github.com/Darmody/fine-uploader/blob/master/client/js/traditional/traditional.xhr.upload.handler.js#L77-L81

I think `isErrorUploadResponse` treat `204` as a success response. So that `parseResponse` will try to parse an empty `responseText`.

https://github.com/FineUploader/fine-uploader/blob/c1b4954f21381dbd664e379abf571c988e2265c0/client/js/traditional/traditional.xhr.upload.handler.js#L103

rnicholus:
The request will be marked as an error the `success` property of the response is "falsy". An empty object falls into that category.

Darmody:
@rnicholus Oh, sorry. I caught your point now.

Darmody:
@rnicholus Is it sufficient now?

rnicholus:
I don't think so. This could have a substantial impact on existing servers that depend on the current behavior. In other words, even if the response code is 2xx, you can signal an error simply by _not_ returning a JSON response with a `success: true` property. This was especially important in IE9 and older, which are still technically supported by Fine Uploader.

I would consider this a breaking change, and, as a result, I can't release this without moving to version 6.0.0. I do have a (stale) v6 branch which I hope to get back to at some point. In that branch, I may very well remove support for all but the most current browsers (support for anything older than IE11 will be gone). There are likely already some open cases that are targeted for 6.0 that should make the response handling logic less awful.

Darmody:
@rnicholus I see, then I will close this one.

rnicholus:
Thanks for you time. Sorry I couldn't merge this in.

PR #1724, opened 2017-01-20
Allow custom error message in S3 signature response

mrdziuban:

This adds logic to the S3 request signer to use the `error` property on the response (if set) as the error message displayed to the user. If the `error` property is not set, the logic will work the same way it currently does. 

On Mac OS X 10.11.6:

- Chrome 55.0.2883.95
- Firefox 50.1.0
- Safari 9.1.2

On Windows 7:

- Internet Explorer 10
- Internet Explorer 11

On Windows 10:

- Edge 38.14393.0.0

Yes

Yes


mrdziuban:
Not sure why tests failed on Travis, but they passed when I ran them locally.

rnicholus:
Not sure what happened - re-running the build results in green. 

This looks good, but two additional requests, if you could:
- [ ] Write a unit test to cover this new behavior
- [ ] Update the documentation to let future integrators know of this possibility

mrdziuban:
@rnicholus thanks, I updated the documentation. I would be happy to add a unit test, but I looked into the existing S3 tests and wasn't able to find a way to tap into the S3 request signer from the test. Let me know if there's an example of this, or what you think the best way to proceed is.

rnicholus:
Thanks for the doc update. There are some existing integration tests for the S3 uploader. [One tests a number of different failure scenarios](https://github.com/FineUploader/fine-uploader/blob/develop/test/unit/s3/chunked-uploads.js#L476). Perhaps you could copy that logic and simplify it a bit to form a new test that focuses on ensuring this results in a failure of the upload. 

I realize this does not test for the presence of the custom message after the failure. Another (possibly better?) option would be to test the `handleSignatureReceived` method in isolation. If you `new qq.s3.RequestSigner({...})`, you should be able to access the `handleSignatureReceived` implementation through [the `onComplete` property on the constructed instance](https://github.com/FineUploader/fine-uploader/blob/develop/client/js/s3/request-signer.js#L471). You could then perhaps mock the parameters as appropriate, and test for the rejected promise value.

mrdziuban:
Thanks again @rnicholus. I was able to add some tests by specifying the `onError` callback to the uploader.

rnicholus:
Awesome, this looks good. Thanks for your work. I'll release this as 5.13.0 sometime soon, possibly even this weekend.

rnicholus:
Just released as 5.13.0.

mrdziuban:
Thanks!

PR #1739, opened 2017-01-31
feature: adding setStatus to the public api list

CoolGoose:

setStatus with DELETE and DELETE_FAILED
extract onDeleteComplete if/else into _handleDeleteSuccess / _handleDeleteFailed ?
written unit tests.

unit tests

yes

yes

### DIFF
--- a/client/js/uploader.basic.api.js
+++ b/client/js/uploader.basic.api.js
@@ -397,6 +397,28 @@
             return this._uploadData.uuidChanged(id, newUuid);
         },
 
+        /**
+         * Expose the internal status of a file id to the public api for manual state changes
+         * @public
+         *
+         * @param {Number} id,
+         * @param {String} newStatus
+         *
+         * @todo Implement the remaining methods
+         */
+        setStatus: function(id, newStatus) {
+            switch (newStatus) {
+                case qq.status.DELETED:
+                    this._handleDeleteSuccess(id);
+                    break;
+                case qq.status.DELETE_FAILED:
+                    this._handleDeleteFailed(id);
+                    break;
+                default:
+                    this.log("Method setStatus called on '" + name + "' not implemented yet for " + newStatus);
+            }
+        },
+
         uploadStoredFiles: function() {
             if (this._storedIds.length === 0) {
                 this._itemError("noFilesError");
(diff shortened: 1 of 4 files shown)


client/js/uploader.basic.api.js
```diff
@@ -1067,6 +1089,35 @@
...
+        },
+
+        _handleDeleteFailed: function(id, xhrOrXdr) {
+            var name = this.getName(id);
+
+            this._uploadData.setStatus(id, qq.status.DELETE_FAILED);
+            this.log("Delete request for '" + name + "' has failed.", "error");
+
+            // Check first if xhrOrXdr is actually passed since manual status changes don't handle this.
+            if (xhrOrXdr) {
```
rnicholus: I think this section should be refactored to always call the `onError` callback, even if `xhrOrXdr` is absent, in which case the error message can simply omit any references to `xhrOrXdr`.

This may work (replaces everything starting w/ line 1108:

```js
// For error reporting, we only have access to the response status if this is not
// an `XDomainRequest`.
if (!xhrOrXdr || xhrOrXdr.withCredentials === undefined) {
    this._options.callbacks.onError(id, name, "Delete request failed", xhrOrXdr);
}
else {
    this._options.callbacks.onError(id, name, "Delete request failed with response code " + xhrOrXdr.status, xhrOrXdr);
}
```

rnicholus (changes requested):
This looks really good! Just a couple final things before we can merge this (in addition to my inline comments):

1. Update documentation - we'll need to add this method to the methods.jmd documentation page. I can do that myself after this is ready to be merged.
2. Write some additional unit tests or increase the level of testing in the new tests you added. For example, I'd like to see a test that confirms the return value of `getNetUploads()` is as expected both when a delete fails and succeeds via the `setStatus` route.
3. We'll want to do some manual testing too. This is something I can handle once we're ready to merge.

CoolGoose:
@rnicholus I've added documentation, getNetUploads options, modified the condition and added a qq.error throw in addition to the console.log . Hopefully i didn't miss anything

test/unit/set-status.js
```diff
@@ -0,0 +1,105 @@
...
+
+        qqtest.downloadFileAsBlob("up.jpg", "image/jpeg").then(function(blob) {
+            fileTestHelper.mockXhr();
+
+            uploader.addFiles({name: "test", blob: blob});
+            uploader.uploadStoredFiles();
+            fileTestHelper.getRequests()[0].respond(201, null, JSON.stringify({success: true}));
+        });
+
+        var uploaderFiles = uploader.getUploads();
```
rnicholus: One more thing - can you please move all test logic into the `then` success/resolved function? Otherwise this can create a race condition where the assertions are executed before the file is added to the uploader. That applies to this test and the one below. Also, make use of the `done` function to signal that the test has actually completed after all assertions have been executed, which is important for async tests such as this one.

rnicholus (changes requested):
Wow, you completed everything. Just one small change to the tests and we'll be ready to run some manual tests before merge.

test/unit/set-status.js
CoolGoose: @rnicholus should be moved now into then, and added done as a param to it and called it by hand after the asserts. I hope it's good :) my mocha testing is somewhere around 0

rnicholus (approved):

rnicholus:
Thanks! After I get some time to do a little manual testing, I'll merge this into develop and publish as 5.14.0-beta2 if all looks well.

rnicholus:
I ended up merging this in via the command line. I did this as I encountered a few unexpected issues during manual testing, which I fixed. The issues were:

- UI does not update when file is deleted via `setStatus()`.
- Error message was unhelpful if file ID is not valid.
- A file could be marked as deleted several times, messing up internal status (such as the net files count).

There were also a couple documentation issues that I missed during my review.

But all is well now. Thanks so much for your help! This has been released as 5.14.0-beta2, now on npm.

PR #1750, opened 2017-02-13
feat(uploader.api.js) added `id` and `uuid` to `showConfirm()`

nalbion:
Added `id` and `uuid` to `showConfirm()`.  I personally have 2 reasons for this:

- I want the $mdDialog to animate from the button which was pressed
- I do virus scanning on upload files and when a user presses "delete" for a file which has been rejected, I don't want the confirm dialog to be shown

N/A

yes

develop

### DIFF
--- a/client/js/uploader.api.js
+++ b/client/js/uploader.api.js
@@ -523,7 +523,7 @@
                 self = this,
                 retVal;
 
-            retVal = this._options.showConfirm(confirmMessage, id, uuid);
+            retVal = this._options.showConfirm(confirmMessage, id);
 
             if (qq.isGenericPromise(retVal)) {
                 retVal.then(function() {
--- a/client/js/uploader.api.js
+++ b/client/js/uploader.api.js
@@ -523,7 +523,7 @@
                 self = this,
                 retVal;
 
-            retVal = this._options.showConfirm(confirmMessage);
+            retVal = this._options.showConfirm(confirmMessage, id, uuid);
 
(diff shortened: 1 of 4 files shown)


rnicholus:
Thanks for the PR! I'll take a closer look as soon as I get a chance, but I would suggest you remove the UUID parameter. It's redundant and can be determined given an ID using the [`getUuid()` method](http://docs.fineuploader.com/branch/master/api/methods.html#getUuid).

client/js/uploader.api.js
```diff
@@ -523,7 +523,7 @@
                 self = this,
                 retVal;
 
-            retVal = this._options.showConfirm(confirmMessage);
+            retVal = this._options.showConfirm(confirmMessage, id, uuid);
```
rnicholus: Please remove `uuid` since this can be obtained via the `getUuid()` API method. Also please update the docs after this change.

rnicholus (changes requested):

nalbion:
Hi @rnicholus, `getUuid()` could be used to get the uuid, but it's available when `showConfirm()` is called, and I need to use the value.

rnicholus:
You may need both parameters, but others may not. For consistency, please stick to the single ID parameter.

client/js/uploader.api.js
nalbion: done

rnicholus (approved):

PR #1796, opened 2017-03-21
Fix s3 pause/continue chunked multiple bug.

JohnTheGray:
Reproduce:
s3.XhrUploadHandler
options:
, chunking: { enabled: true, concurrent: { enabled: true } }
, maxConnections: 3
, multiple: true
, resume: { enabled: true}

Chrome: Version 56.0.2924.87 (64-bit) (also on Edge)

Begin uploading a large file. Pause and continue the upload a few times
and notice that immediately or eventually (it's sporadic) the upload will
be paused but chunks will still upload in the background.

Solution: the paused state is not cleared on upload continue. initHeaders
promise success seems to be called even though the XHR is aborted. Added
file status pause check here to prevent an XHR PUT when file hsa been
paused.

### DIFF
--- a/client/js/upload-handler/xhr.upload.handler.js
+++ b/client/js/upload-handler/xhr.upload.handler.js
@@ -118,8 +118,7 @@ qq.XhrUploadHandler = function(spec) {
                     name: uploadData.name,
                     remaining: uploadData.chunking.remaining,
                     size: uploadData.size,
-                    uuid: uploadData.uuid,
-                    loaded: uploadData.loaded
+                    uuid: uploadData.uuid
                 };
 
                 if (uploadData.key) {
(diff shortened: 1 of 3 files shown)


client/js/upload-handler/xhr.upload.handler.js
```diff
@@ -118,7 +118,8 @@ qq.XhrUploadHandler = function(spec) {
                     name: uploadData.name,
                     remaining: uploadData.chunking.remaining,
                     size: uploadData.size,
-                    uuid: uploadData.uuid
+                    uuid: uploadData.uuid,
+                    loaded: uploadData.loaded
```
rnicholus: This is not necessary. The number of bytes remaining is already something we can calculate based on the chunk size and the remaining chunks. 

Also, please do not combine unrelated changes into a single pull request. I'd very much like to keep changes simple and focused.

Issue #1802, opened 2017-03-31
Initial/canned file's thumbnailUrl is not available until a few ticks after qq.status.UPLOAD_SUCCESSFUL


rnicholus:
Discovered in React Fine Uploader at FineUploader/react-fine-uploader#91. To properly file, the internal state must be completely updated (which includes the `thumbnailUrl`) with the data from the parsed initial file record _before_ `qq.status.UPLOAD_SUCCESSFUL` is triggered. The issue can be seen in the [`_addCannedFile` method](https://github.com/FineUploader/fine-uploader/blob/5.14.1/client/js/uploader.basic.api.js#L444), which will have to be adjusted to ensure the event/status change is not broadcast until all internal state is updated.

PR #1803, opened 2017-03-31
Add support for S3 SSE using own KMS key

vladgolubev:

Huge thanks for the brilliant library! 💯 

I found that Fine Uploader supports uploading to S3 with SSE. That's great! But it lacks some not so popular feature -  [encryption with custom KMS Keys](http://docs.aws.amazon.com/AmazonS3/latest/dev/KMSUsingRESTAPI.html)  which I need.

Fortunately, it was pretty straightforward to patch the lib for my needs.

But this is not a proper long-term solution. So that's why I propose this PR to raise a discussion about adding this feature to Fine Uploader. Treat this PR as Proof of Concept as it was done in "Search & Replace" mode and doesn't pretend to be merged.

Chrome 57.0.2987.110 (64-bit)

I'm not sure you'll even decide to include this feature into Fine Uploader

But I added some unit tests (which fail for some reason)

## Results

So I did a simple test with small (10KB) and big (53MB) file. And looks like it was encrypted properly:

image

What do you think @rnicholus  ?


rnicholus:
Thank you for your work on this, and for bringing this to my attention. I find that I learn more about S3 through user-contributed issues and PRs in this project than anywhere else. This is interesting, and appears to be a low-risk addition, so I don't see an immediate reason to reject this outright. As you said, the tests _are_ failing though. One other note, SSE w/ KMS is only supported when using v4 signatures, so that would require some changes to your PR I believe.

Currently, my priority is to release Fine Uploader 5.14.2 once #1802 is solved, which is needed to push out the first stable release of React Fine Uploader. I'm also about to start a new job, and I'm the only consistent maintainer of this large suite of projects. So, it may be a while before I can take some time to code review and offer assistance here. But here's what I think is needed to push this forward:

1. Updates to reflect that this feature is only supported w/ v4 signatures.
2. Passing tests.
3. Updates to the documentation (in this case, the S3 feature page + options).
4. A full code review by me, which may result in more requested changes.

vladgolubev:
Whoa! 

Cool solution @nadeesha!

PR #1818, opened 2017-04-13
test(package): lock pica to v2.0.8

Novex:
Pica very recently released v3.0.0, and appear to have renamed `pica.resizeCanvas` to something else. This breaks the test suite, but not the build.

This commit locks pica to the last working version, 2.0.8 in order to make the test suite pass again.

Verified that the test suite now passes on Windows 8.1

No, this is to make existing tests work again.


rnicholus:
Good deal, thanks!

PR #1828, opened 2017-04-26
corrected case of `client/commonJs`

nalbion:
`make setup-dist` failed because `client/commonjs` could not be found

Linux

No


rnicholus:
Does look like a legit bug, but I haven't noticed as I currently manually build/public from one of my dev machines, which are all Macs w/ HFS+ filesystems.

PR #1836, opened 2017-05-18
Added Polyfill for CustomEvent in IE >= 9

thejanasatan:
Changed the way that dnd.js creates the HideZone event. IE 10 throws an exception when used with an if-else. 

IE >= 9

Nop. :|


rnicholus:
> IE 10 throws an exception when used with an if-else.

Can you expand on this? Yours is the first such report, and this code has been live for quite a long time.

thejanasatan:
Here's a JSFiddle I wrote to test this. 
https://jsfiddle.net/bLenv0w1/2/

The first chunk of code with if-else throws an error in IE 11. But the second chunk does not (without if-else.

The if-else here doesn't work. Check: https://github.com/alexgibson/tap.js/pull/26

thejanasatan:
@rnicholus bump

rnicholus:
sorry - i don't have a lot of time to keep up this lib lately, though I am working on some big changes in https://github.com/FineUploader/fine-uploader/tree/feature/flexible-upload-support. At some point, I don't expect to support anything older than IE11 anyway...

Issue #1838, opened 2017-05-18
Typescript Browser ReferenceError: qq is not defined


jclangst:
## Type of issue
- [x] Bug report
- [ ] Feature request
- [ ] Support request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

<details>
<summary>Bug Report</summary>

#### Fine Uploader version
5.14.2

#### Browsers where the bug is reproducible
Chrome 58

#### Operating systems where the bug is reproducible
Ubuntu 17.04

#### Exact steps required to reproduce the issue
Note: This occurs in a Typescript Angular project being bundled by Webpack with the awesome-typescript-loader.

My `tsconfig.json`:

```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2015",
    "sourceMap": true,
    "outDir": "build",
    "moduleResolution": "node",
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "noImplicitAny": false,
    "suppressImplicitAnyIndexErrors": true,
    "lib": ["es2015", "dom"],
    "allowJs": false
  },
  "exclude": [
    "node_modules",
    "app",
    "public"
  ],
  "awesomeTypescripteLoaderOptions": {

  }
}
```

Error was discovered using the below file snippet, but can also be reproduced using the first 26 lines of `fine-uploader.tests.ts'.

The error occurs at `new qq.FineUploader(...)`.

#### All relevant Fine Uploader-related code that you have written
```
/// <reference types="fine-uploader" />
...
export class AboutComponent implements OnInit{


    editorDOM: HTMLElement;
    editorOutputDOM: HTMLElement;
    titleInput: HTMLElement;
    dateInput: HTMLElement;
    uploaderInput: HTMLElement;

    editor: ace.Editor;
    taggle: Taggle;
    uploader: FineUploader.qq;

    post: Post;

   ...
    getDOM(): void {
        this.editorDOM = document.getElementById("md-editor");
        this.editorOutputDOM = document.getElementById("md-editor-output");
        this.titleInput = document.getElementById("post-title");
        this.dateInput = document.getElementById("post-date");
        this.uploaderInput = document.getElementById("fine-uploader");
    }
   ...

    initUploader(): void {

        this.uploader = new qq.FineUploader({
            element: this.uploaderInput,
            validation: {
                allowedExtensions: ['png', 'jpeg', 'jpg', 'svg']
            },
            blobs: {
[...]

PR #1840, opened 2017-05-19
Updated Typescript to support canonical imports

fullykubed:
Solves issue explained in #1838. Adds support for the canonical Typescript imports. Matches the existing import syntax shown in the [import description page](https://docs.fineuploader.com/features/modules.html) except that the `qq` can be dropped as it now properly knows that `qq` does not exist on the global namespace (due to the UMD modular design).

Typescript example:

```
// use Fine Uploader core (non-ui) for traditional endpoints
import qq from 'fine-uploader/lib/core'

const uploader = new qq.FineUploaderBasic({...})
```
**becomes**
```
// use Fine Uploader core (non-ui) for traditional endpoints
import {FineUploaderBasic} from 'fine-uploader/lib/core'

let uploader = new FineUploaderBasic({...})
```

Couple of notes:
- To avoid confusion in the future, I would add some typescript instructions into the  [import description page](https://docs.fineuploader.com/features/modules.html).
- Since Typescript gets rid of `qq` on the global namespace, all of the utility functions that were on there are not visible to Typescript users. I assumed this was fine as it appears they are only being used internally.
- It doesn't look like `fine-uploader.test.ts` hooks up to any runtime tests which I think was the root cause for this issue originally. I did not want to mess up your test flow (and `npm test` kept failing on the clean repo for me), but they should be fairly trivial to make.

Chome 58
Typescript 2.3

Updated the `fine-uploader.test.ts` file to show modularized Typescript syntax

### DIFF
--- a/client/typescript/fine-uploader.d.ts
+++ b/client/typescript/fine-uploader.d.ts
@@ -1,17 +1,596 @@
 // Type definitions for FineUploader 5.x.x
 // Project: http://fineuploader.com/
 // Definitions by: Sukhdeep Singh <https://github.com/SinghSukhdeep>
+//
+// declare module "qq"{
+//     export = qq;
+// }
 
-/**
- * The FineUploader namespace contains all the methods, options, events and types
- */
-declare namespace FineUploader {
 
-    /* ========================================================== CORE & UI ===================================================================== */
+declare module "fine-uploader/lib/core" {
+
+
+    export class FineUploaderBasic {
+
+        /**
+         * The FineUploader Core only constructor
+         */
+        constructor (fineuploaderOptions?: CoreOptions);
+
+        /**
+         * FineUploader's Promise implementation
+         */
+        Promise(): void;
+
+        /**
+         * Submit one or more files to the uploader
+         *
+         * @param any[] files : An array of `File`s, `<input>`s, `Blob`s, `BlobWrapper` objects, `<canvas>`es, or `CanvasWrapper` objects. You may also pass in a `FileList`.
+         * @param any params : A set of parameters to send with the file to be added
+         * @param string endpoint : The endpoint to send this file to
+         */
+        addFiles(files: File[]
+                     | HTMLInputElement[]
+                     | Blob[]
+                     | BlobWrapper
+                     | HTMLCanvasElement[]
+                     | CanvasWrapper
+                     | FileList, params?: any, endpoint?: string): void;
+
+        /**
+         * Submit one or more canned/initial files to the uploader
+         *
+         * @param any[] initialFiles : An array of objects that describe files already on the server
+         */
+        addInitialFiles(initialFiles: any[]): void;
+
+        /**
+         * Cancel the queued or currently uploading item which corresponds to the id
+         *
+         * @param number id : The file's id
+         */
+        cancel(id: number): void;
+
+        /**
+         * Cancels all queued or currently uploading items
+         */
+        cancelAll(): void;
+
+        /**
+         * Clears the internal list of stored items. Only applies when autoUpload is false
+         */
+        clearStoredFiles(): void;
+
+        /**
+         * Attempts to continue a paused upload
+         *
+         * @param number id : A file id
+         * @returns boolean : `true` if attempt was successful.
+         */
+        continueUpload(id: number): boolean;
+
+        /**
+         * Send a delete request to the server for the corresponding file id
+         *
+         * @param number id : The file's id
+         */
+        deleteFile(id: number): void;
+
+        /**
+         * Draws a thumbnail
+         *
+         * @param number id : The id of the image file
+         * @param HTMLElement targetContainer : The element where the image preview will be drawn. Must be either an <img> or <canvas> element
+         * @param number maxSize : The maximum dimensions (for width and height) you will allow this image to scale to
+         * @param boolean fromServer : true if the image data will come as a response from the server rather than be generated client-side
+         * @param CustomResizerCallBack customResizer : Ignored if the current browser does not support image previews.
+         *                                              If you want to use an alternate library to resize the image, you must contribute a function for this option that returns a `Promise`.
+         *                                              Once the resize is complete, your promise must be fulfilled.
+         *                                              You may, of course, reject your returned `Promise` is the resize fails in some way.
+         * @returns Promise: Fulfilled by passing the container back into the success callback after the thumbnail has been rendered.
+         *                   If the thumbnail cannot be rendered, failure callbacks will be invoked instead, passing an object with `container` and `error` properties.
+         */
+        drawThumbnail(id: number, targetContainer: HTMLElement, maxSize?: number, fromServer?: boolean, customResizer?: CustomResizerCallBack): PromiseOptions;
+
+        /**
+         * Returns the button container element associated with a file
+         *
+         * @param number id : The file id
+         * @returns HTMLElement : The button container element associated with a file, or `undefined` if the file was not submitted via a Fine Uploader controlled upload button.
+         */
+        getButton(id: number): HTMLElement;
+
+        /**
+         * Returns the file identified by the id. File API browsers only
+         *
+         * @param number id : The file id
+         * @returns File | Blob : A `File` or `Blob` object
+         */
+        getFile(id: number): File | Blob;
+
+        /**
+         * Returns the endpoint associated with a particular file, or the current catch-all endpoint for all files (if no ID is specified).
+         *
+         * @param number id : The ID of the associated file
+         * @return string | string[] : endpoint associated with a particular file, or the current catch-all endpoint for all files (if no ID is specified).
+         */
+        getEndpoint(id?: number): string | string[];
+
+        /**
+         * Returns the number of items that are either currently uploading or waiting for an available connection (`qq.status.QUEUED`).
+         *
+         * If called inside of a cancel event handler, then this method will return a value that includes the upload associated with the cancel event handler.
+         * This is because the upload will not be canceled until the event handler returns.
+         *
+         * @returns number : The number of items that are currently uploading or queued
+         */
+        getInProgress(): number;
+
+        /**
+         * Returns the name of the file with the associated id
+         *
+         * @param number id : The file id
+         * @returns string : Returns the name of the file identified by the id.
+         */
+        getName(id: number): string;
+
+        /**
+         * Get the number of items that have been successfully uploaded and have not been deleted
+         *
+         * @returns number : The number of items that have been successfully uploaded and not deleted
+         */
+        getNetUploads(): number;
+
+        /**
+         * Get the ID of the parent file for this scaled file
+         *
+         * @param number scaledFileId : The ID of a scaled image file
+         * @returns number : Returns the ID of the scaled image's parent file. `null` if this is not a scaled image or a parent cannot be located
+         */
+        getParentId(scaledFileId: number): number;
+
+        /**
+         * Returns the number of remaining allowed items that may be submitted for upload based on `validation.itemLimit`.
+         */
+        getRemainingAllowedItems(): number;
+
+        /**
+         * Returns an array of potentially resumable items
+         *
+         * @returns ResumableFileObject[] : An array of resumable items
+         */
+        getResumableFilesData(): ResumableFileObject[] | ResumableFileObject;
+
+        /**
+         * Returns the size of the item with the associated id
+         *
+         * @param number id : The file id
+         * @returns number : The size of the file with the corresponding id
+         */
+        getSize(id: number): number;
+
+        /**
+         * Return information about all the items that have been submitted to the uploader
+         *
+         * @param UploadFilter filter : An object which indicates which keys and values must be present in an upload to be returned
+         * @return FoundUploadItems | FoundUploadItems [] : A list of items or a single item that has been filtered/found.
+         *                                                  This returns an array only when there is a potential for the operation to return more than one file in the result set.
+         *                                                  This excludes queries for a specific, single ID or UUID. All other queries will return an array
+         */
+        getUploads(filter?: UploadFilter): FoundUploadItems | FoundUploadItems[];
+
+        /**
+         * Returns the UUID of the item with the associated id
+         *
+         * @param number id : The file id
+         * @returns string : A level 4 UUID which identifies the corresponding file
+         */
+        getUuid(id: number): string;
+
+        /**
+         * Output a message to the console, if possible
+         *
+         * @param string message : The message to print
+         * @param string level : The level to output the message at
+         */
+        log(message: string, level?: string): void;
+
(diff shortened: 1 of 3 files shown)


SinghSukhdeep:
I see that the def file has changed considerably. It will take me some time to get through this.

jclangst:
As a note to aid in your review, I fortunately didn't change any of the
definitions of any of the functions. I just exposed the js classes as
Typescript classes, and encapsulated into them the appropriate functions
(based on the online API).

This was done since Typescript users won't have access to the qq global
variable after they transpile their ts to js -- Typescript users will
usually target a module system like AMD, CommonJS, or ES6 and the UMD
design of the project will then not expose everything under a qq variable
but rather on an export object. That's why the Typescript docs recommend
using the module template (
https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html
).

I am not sure why having it the current way would have allowed rollup to
work with Typescript as it seems it should have still broken it. If you
want to use these new Typescript definitions with Rollup all you'd need to
do is change the module option of the tsconfig.json file to "ES6." While I
haven't tested it, this will give you the import/export syntax that should
allow rollup to work.

What I do know, however, is that the current definitions will trick the
Typescript compiler into allowing code that will lead to runtime errors.

Let me know if I can help in any way.

Jack

SinghSukhdeep:
Current structure is set in a way that you have to obviously include fine-uploader JS file globally in your index.html and provide support to be able to use the same syntax for TS projects.
And hence we wouldn't need Rollup or any other module bundler/loader to load the library, that's how it is working right now.

I totally understand what you are saying or trying to implement here. But I'm afraid it's not going to work with Rollup as of yet.
I haven't tried your changes with Rollup yet, I will soon.

Also just a note that Rollup isn't my obsession, it's what is recommended by Angular team itself and in all their documentation. Meaning that there are lots of people out there using exactly that myself included.

jclangst:
Ah, I see what you are saying: there should be a both an import statement in the Typescript and a global include in the index.html file. That would certainly work, but it would be nice to just have to import the module once in our Typescript projects which is the functionality that the new .d.ts file provides and is recommended by the Typescript crew. Also, this would work much better with bundlers (including Rollup, I think).

Also, Rollup is certainly a wonderful utility (just haven't used it personally), and I did not mean to imply in any way that you should not love it and support it. What I am trying to say is that when transpiling Typescript, it will produce javascript that does not expose the global `qq` variable. Thus, you will get a runtime error regardless of whether you use Rollup or not -- unless, as you pointed out, you do a second include of the underlying javascript at a different point in your html code.

If you think I am wrong (and I certainly could be!), could you provide an example of how you are using Typescript with Rollup and not getting resulting runtime errors? I might be missing something.

I think that this whole issue is probably a result of the [import documentation page](https://docs.fineuploader.com/branch/master/features/modules.html). It says that ES6 import syntax can be used with languages like Typescript and utilities like Webpack and gives several examples -- none of which would work in Typescript. That's what led me to develop the new Typescript definition file that now follows the Typescript's team's recommendations for UMD projects and will work out of the box with the instructions in the documentation.

rnicholus:
The document you are referring to was written by me, before TS support was added. And I don't personally use TS, so it's likely we need dedicated TS docs too.

jclangst:
No worries! Definitely hard to keep up with everything popping up every other month. Fortunately, the original instructions should be 100% fine -- we just need to have the .d.ts file treating the modules as modules rather than using the global namespace (which won't have anything on it since fine-uploader uses UMD).

SinghSukhdeep:
I haven't forgotten about this. Haven't had any time to work on this.
I'm hoping to dedicate some time this weekend.

SinghSukhdeep:
Did some testing and I've put together a repo [here](https://github.com/SinghSukhdeep/fine-uploader-ts) where all the changes are seems to be working with SystemJS and Rollup (Thanks to improvements rollup's commonJS plugin).

I need to do some more testing on a real Angular 4 & TypeScript project and then we can go further from there.

I am optimistic that this will be a great benefit to Angular & TypeScript based projects, but is also a major syntax change for the existing users who are using globally scoped FineUploader JS with current TS definitions. 

Also we need to document this change properly and maybe have a dedicated page or section for Angular 2+ and TypeScript integration.

jclangst:
Awesome!

Let me know if I can help with anything.

My two cents would be just having an example import statement for TS on the
documentation for importing would be enough. Everything else should
function the same way.

Best,
Jack

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+             * You must return a promise. If your attempt to refresh the temporary credentials is successful, you must fulfill the promise via the success method, passing the new credentials object.
+             * Otherwise, call failure with a descriptive reason.
+             */
+            onCredentialsExpired?: OnCredentialsExpired;
+        }
+
+        /**
+         * S3UIOptions
+         */
+        export interface S3UIOptions extends UIOptions {
```
SinghSukhdeep: @jleider this interface must also extend S3CoreOptions

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+            /**
+             * AzureUploadSuccessOptions
+             */
+            uploadSuccess?: AzureUploadSuccessOptions;
+        }
+
+        /**
+         * AzureUIOptions
+         */
+        export interface AzureUIOptions extends UIOptions {
```
SinghSukhdeep: @jleider this interface must also extend AzureUIOptions

singhjusraj (changes requested):

SinghSukhdeep:
I think it'll be better now if we divide the modules into individual files for each and have one index.ts exporting everything.

client/typescript/fine-uploader.d.ts
jclangst: Resolved:

https://github.com/FineUploader/fine-uploader/pull/1840/commits/e95631cc61fd459756aebcbf6d05d33c1c2a91d2

client/typescript/fine-uploader.d.ts
jclangst: Resolved:

https://github.com/FineUploader/fine-uploader/pull/1840/commits/e95631cc61fd459756aebcbf6d05d33c1c2a91d2

jclangst:
Yes. We can split the ambient module declarations into separate files (even splitting separate module elements into their own modules if we want to break it up even further); however, these new modules must have different names than those names in the final index.d.ts file (which need to remain as they are).

At what level do you want to split things into separate files? Current module level? Class/interface level?

SinghSukhdeep:
I think splitting by current module level should be okay for now, we can always split it further if needed

SinghSukhdeep:
In addition to 

```
'fine-uploader/lib/core'
'fine-uploader'
'fine-uploader/lib/s3'
'fine-uploader/lib/azure'
```

also need to declare the `'fine-uploader/lib/all'` module which would essentially have all the flavours

irbian:
Hi!

I have read some of your threads regarding the integration with angular 2. Is this one the most up to date? We are very interested on integrate this library with our project because we need the IE9 support

SinghSukhdeep:
@irbian as you can see in this PR we're in the middle of transforming our TypeScript support.
If you want to use FineUploader in your project at the moment, look at this [file](https://github.com/FineUploader/fine-uploader/blob/develop/client/typescript/fine-uploader.test.ts) for syntax.
However it will change for better after we merge this PR.

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+             * @default `'qqfilename'`
+             */
+            filenameParam?: string;
+        }
+
+        /**
+         * type for Azure's customHeaders function
+         */
+        export interface AzureCustomHeaderFunction {
+            constructor(id: number);
```
SinghSukhdeep: @jclangst is there any reason to change this interface's declaration?
Before it was `(id: number): void;` and now it is `constructor(id: number);`

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
-         * @param string endpoint : An endpoint that Fine Uploader should POST to when a file has been successfully uploaded to Azure Blob Storage
-         * @param number id : An ID corresponding to a file
+         * AzureSignatureOptions
          */
-        setUploadSuccessEndpoint(endpoint: string, id?: number): void;
+        export interface AzureSignatureOptions {
+            /**
+             * Additional headers sent along with each signature request.
+             *
+             * If you  a function as the value, the associated file's ID will be passed to your function when it is invoked
```
SinghSukhdeep: @jclangst I guess since you did replace all for all occurrences of word `declare`, it has led to sentences like those. 
see `declare` missing between `If you` and the rest of the sentence

client/typescript/fine-uploader.d.ts
```diff
@@ -1,2975 +1,2306 @@
...
-    interface OnError {
+        mandatory?: boolean;
         /**
-         * @param number id : The current file's id
-         * @param string name : The current file's name
-         * @param string errorReason : The reason for the current error
-         * @param XMLHttpRequest xhr : The object used to make the request
+         * The maximum size of each chunk, in bytes
+         *
+         * @default `200000
```
SinghSukhdeep: @jclangst check this sentence as well, due to replacement of declare

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+        }
+
+        /**
+         * S3SignatureOptions
+         */
+        export interface S3SignatureOptions {
+            /**
+             * Additional headers sent along with each signature request.
+             *
+             * If you  a function as the value, the associated file's ID will be passed to your function when it is invoked
```
SinghSukhdeep: @jclangst same problem caused by replacing declare here as well

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+         */
+        export interface HostFunction {
+            constructor(id: number);
+        }
+
+        /**
+         * type for S3's key object property
+         */
+        export interface KeyFunction {
+            constructor(id: number);
```
SinghSukhdeep: @jclangst this function and 3 above it has declaration like this `(id: number): PromiseOptions | string;` before.
Any reason for the change?

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+             * @default `{}`
+             */
+            params?: any;
+        }
+
+        /**
+         * type for S3's customHeaders function
+         */
+        export interface S3CustomHeaderFunction {
+            constructor(id: number);
```
SinghSukhdeep: @jclangst this function declaration has been changed as well?

client/typescript/fine-uploader.d.ts
```diff
@@ -2984,397 +2315,1240 @@ declare namespace FineUploader {
...
+             * @default `'custom'`
+             */
+            mode?: string;
+        }
+
+        /**
+         * onCredentialsExpired function type
+         */
+        export interface OnCredentialsExpired {
+            constructor();
```
SinghSukhdeep: @jclangst this function declaration has been changed as well

client/typescript/fine-uploader.d.ts
SinghSukhdeep: @jclangst I see lot of function declarations have been changed to use constructors.
Is there a specific problem with previous declarations?

singhjusraj (commented):
Can you please address those things

SinghSukhdeep:
@jclangst  Are you still pursuing this PR?

If not, the things that I mentioned in my last review, I'll have to revert those back to what it was before and then merge this. I'm hoping to have this released very soon as many users are requesting this

singhjusraj (approved):

SinghSukhdeep:
@rnicholus Can you please review the documentation changes from this PR and also `qq` and its related functions are no longer available for TS users, so slight syntax change is required for all TS users.

I have updated the documentation to reflect how library should be imported in TS, is there anything else we need to do before we merge this?

rnicholus:
Are these the breaking changes you mentioned to me earlier? If so, looks like this only affects Typescript users. Is this correct?

Docs look fine to me. I generally only use `let` if I know a variable needs to be changed, which means I default to `const` instead. But that's really a nitpick here. 

Feel free to merge and release. Please do familiarize yourself and follow the release (and also the merging) guidelines at https://github.com/FineUploader/fine-uploader/wiki/Maintainers-guide first.

SinghSukhdeep:
Yes this only affects TypeScript users, no one else.
`let` is mostly used in TS instead of using `var` directly, but it actually compiles down to `var`, so don't worry about it.

I've read the maintainers guide before, 
If I understand correctly, following are the steps:

1. Update the version number in package.json and version.js
2. Then do squash and merge this PR into develop branch
3. Then merge develop into master
4. Run `publish` Makefile recipe from master branch, which would publish to npm.
5. Create a new Release in the Releases section following the instructions

Step 6 in the Releasing steps about `customize.js` is bit unclear to me.

And what should be the next version number 5.14.6 or 5.15.0 since this is major change for TS users. I see you already have few other PR's on 5.15.0 so I'm guessing it'll probably be 5.14.6.

rnicholus:
I'm fine with let, but my point is const is usually more appropriate. Both compile down to var, but const gives you additional safety checking at build time. 

Let's continue to follow semver when deciding version number. If this fixes a bug only, patch version increment. If it's more if a feature, minor version increment. Can you comment on the scope of breaking changes for typescript users? How substantial are the breaking changes?

SinghSukhdeep:
It's as described in the PR description. Changes to the import statement as outlined in the doc update and all of the utility functions being used like `qq.functionName` won't be available.

You told me once that you are planning to remove those in future releases, so I guess that's fine. Besides TS users really don't need those functions.

`import qq from 'fine-uploader/lib/all'` also isn't available because the global namespace `qq` isn't there.

So essentially in order to use all flavour together, separate import is required for each flavor.

Other than those changes, there isn't any change in the syntax other than what's described in the PR description.

So in order to release this as v5.15.0, need to move the PR's currently under this tag to another milestone

quasi13:
@SinghSukhdeep I noticed 2 things regarding the TS canonical imports:

1. 'allowEmpty' seems to be absent from interface ValidationOptions
2. I have not been able to successfully instantiate a new PromiseOptions as done here: https://github.com/FineUploader/fine-uploader/blob/5.15.3/client/typescript/fine-uploader.test.ts#L94

Possibly, both of these are misunderstandings on my part.

Issue #1841, opened 2017-05-22
Using none standard file extensions e.g. '*.bam'


keiranmraine:
## Type of issue
- [ ] Bug report
- [ ] Feature request
- [x] Support request

## Uploader type
- [ ] Traditional
- [x] S3 (Ceph)
- [ ] Azure

<details>
<summary>Support Request</summary>

#### Fine Uploader version
5.14.2

#### Question
We are attempting to use the tool to upload bioinformatics file types such as [.bam](http://bioportal.bioontology.org/ontologies/EDAM?p=classes&conceptid=format_2572) and the submission fails indicating:

```
Received response status 403...Policy missing condition: Content-Type
```

I understand that is due to an inability to identify the mime-type, but can't see where I can setup a mapping for `.bam`.
</details>

rnicholus:
code please

keiranmraine:
I have managed to isolate that this is only a problem for files that don't need to be chunked.

e.g.
`1mb.rnd` fails
`50md.rnd` works

Just the standard S3 example + php server with this as the in html config:

```
var uploader = new qq.s3.FineUploader({
            debug: true,
            element: document.getElementById('fine-uploader'),
            cors: {
              expected: true
            },
            objectProperties: {
              bucket: 'test',
              host: 'our.s3.url'
            },
            request: {
                endpoint: 'https://test.our.s3.url',
                accessKey: 'XXXXXXXX'
            },
            signature: {
                endpoint: "/vendor/fineuploader/php-s3-server/endpoint-cors.php"
            },
            uploadSuccess: {
                endpoint: "/vendor/fineuploader/php-s3-server/endpoint-cors.php?success"
            },
            iframeSupport: {
                localBlankPagePath: "success.html"
            },
            chunking: {
                enabled: true,
                concurrent: {
                    enabled: true
                }
            },
            resume: {
                enabled: true
            },
            retry: {
                enableAuto: true,
                showButton: true
            },
            deleteFile: {
                enabled: true,
                endpoint: "/vendor/fineuploader/php-s3-server/endpoint-cors.php"
            }
        });
```

And minimal cors
```
## CONFIGURE CORS ON BUCKET:
<?xml version="1.0" encoding="UTF-8"?>
<CORSConfiguration>
    <CORSRule>
        <AllowedOrigin>*</AllowedOrigin>
        <AllowedMethod>POST</AllowedMethod>
        <AllowedMethod>PUT</AllowedMethod>
[...]

PR #1846, opened 2017-05-24
set default content-type for uploads

davebiffuk:
If the browser can't figure out a content-type, set a generic one, to avoid policy errors on missing content-type.

This fixes https://github.com/FineUploader/fine-uploader/issues/1841

Tested with Chrome 58 on Linux and Ceph/radosgw/s3 10.2.3

Don't know how :( but willing to learn...

### DIFF
--- a/client/js/s3/util.js
+++ b/client/js/s3/util.js
@@ -136,6 +136,9 @@ qq.s3.util = qq.s3.util || (function() {
             if (type) {
                 conditions.push({"Content-Type": type});
             }
+            else {
+                conditions.push({"Content-Type": "application/octet-stream"});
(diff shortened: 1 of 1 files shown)


rnicholus:
Really having trouble believing that this is necessary. The S3 upload code has been around for _years_, and this is the first time the issue has surfaced.

davebiffuk:
Thanks for the feedback. I've confirmed that this change isn't necessary with AWS S3. I'll pursue it with the Ceph/radosgw people. Would you be amenable to carrying this as a disabled-by-default feature, e.g. "enable_sending_default_content_type"?

PR #1850, opened 2017-05-27
Fix extraButtons typescript define not correct

maivantan1992:
Brief description of the changes

extraButtons in typescript define not correct. Due to document, this option allow an array of ExtraButtonsOptions

What browsers and operating systems have you tested these changes on?

NA
Have you written unit tests? If not, explain why.

NA


SinghSukhdeep:
@maivantan1992  thanks for your contribution

Issue #1862, opened 2017-06-20
Firefox 54 drag area flickers .qq-upload-drop-area-active


SpazzMarticus:
## Type of issue
- [x] Bug report
- [ ] Feature request
- [ ] Support request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

<details>
<summary>Bug Report</summary>

#### Fine Uploader version
5.11.7 on https://fineuploader.com/demos.html
5.14.4 on custom page.

#### Browsers where the bug is reproducible
Firefox 54

#### Operating systems where the bug is reproducible
Windows 8.1

#### Exact steps required to reproduce the issue
Drag file over drop area.

#### All relevant Fine Uploader-related code that you have written
none

#### Your Fine Uploader template markup (if using Fine Uploader UI and the issue is UI-related)
default

#### Detailed explanation of the problem
Dragging file over drop area adds and removes `.qq-upload-drop-area-active` continously.

![2017-06-20_14-42-37](https://user-images.githubusercontent.com/5716457/27333832-909e8a04-55c7-11e7-9347-152d319d4bed.gif)

</details>

cntrytwist:
I can confirm the same behavior when using Mac OS and Firefox 54.

seisnet:
me too

SpazzMarticus:
The problem is caused by [leavingDocumentOut](https://github.com/FineUploader/fine-uploader/blob/5.14.4/client/js/dnd.js#L218-L220). Once you remove this part

```js
if (qq.firefox()) {
    return !e.relatedTarget;
}
```
it works in Firefox 54 as expected, but I guess in older Firefox versions there will be problems.

cubiclesoft:
Ditto.  The flickering issue shows up in the demos in Firefox:

https://fineuploader.com/demos.html

TonyBogdanov:
+1

maximilianschmid:
+1  flickering Firefox 55.0.3 (64-Bit) on MacOS 10.11.6

AyrtonRicardo:
+1 flickering Firefox 54.0 (64-bit) Linux

simshaun:
Happens in Win/Firefox 56 as well.

alphestdev:
Very disturbing for all Firefox users. Still occurs in FF 56, FineUploader 5.15.0.

Removing 3 lines as suggested by SpassMarticus works for full js. It's more complicated removing this safely from a minified version.

Please fix.

[...]

PR #1871, opened 2017-06-30
Fixed Drag'n'Drop for Firefox 54

SpazzMarticus:
Fix for #1862 
Reordered checks in `leavingDocumentOut` in order to try use `e.x` and `e.y`, before falling back on `e.relatedTarget`.

Windows 8.1 x64 - Worked with Firefox 47 **and** 54. 
_I don't know if this has anything to do with it: https://bugzilla.mozilla.org/show_bug.cgi?id=962251)_
(Works in Chrome 59 too.)

Did not find any unit tests for drag and drop.

### DIFF
--- a/client/js/dnd.js
+++ b/client/js/dnd.js
@@ -215,15 +215,15 @@ qq.DragAndDrop = function(o) {
     // * IE10+: If the file is dragged out of the window too quickly, IE does not set the expected values of the
     //          event's X & Y properties.
     function leavingDocumentOut(e) {
-        if (qq.firefox()) {
-            return !e.relatedTarget;
-        }
-
         if (qq.safari()) {
             return e.x < 0 || e.y < 0;
         }
 
-        return e.x === 0 && e.y === 0;
+        if ("x" in e && "y" in e) {
+            return e.x === 0 && e.y === 0;
+        }
+
+        return !e.relatedTarget;
     }
 
     function setupDragDrop() {


cubiclesoft:
A change in drag-and-drop support - one of the core features of a file uploader - needs to be tested widely across all browsers with as many versions as is reasonably possible.  This PR seems like a rather significant change to just accept into mainline with such minimal testing.  A less invasive approach would be to isolate just Firefox using the existing `qq.firefox()` check.

maximilianschmid:
Tested patch successfully on 

- IE11 Win732bit
- FF 55 Mac
- FF 55 32bit Win732bit
- FF 30 Win732bit
- Safari 10.1
- Google Chrome 60  Mac & Win

rnicholus:
I’d like some explanation regarding the details of this fix. It’s not clear to me why this fixes the issue, which I have not verified myself.

SpazzMarticus:
My fix does not work correctly any more in Firefox 56 (x64, Win) any more, because when leaving the browser window I get this error: [`Permission denied to access property "x"`](https://github.com/SpazzMarticus/fine-uploader/blob/8a809af1b2d4ad0e8d614406e5b33f2975e1c238/client/js/dnd.js#L222)
It solves the anoying flickering problem, but does not hide the Drop-Zone when leaving the browser window.

Also in Firefox 56 (x64, Ubuntu 17) the `x` and `y` are not set to 0 or negative values when leaving the browser window.

I hope you can fix this problem!

Issue #1872, opened 2017-06-30
CORS form response, it is not necessary include the UUID


ICELI:
## Type of issue
- [ ] Bug report
- [x] Feature request
- [ ] Support request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

#### Feature request details 
[CORS DOCS](https://docs.fineuploader.com/branch/master/features/CORS.html) Handling iframe CORS upload requests server-side
` and the valid JSON portion must include the UUID of the associated file. `
why? 
In other cases uuid is not necessary, but it's need the UUID when CORS in the form response JSON.
[form.upload.handler.js#L114](https://github.com/FineUploader/fine-uploader/blob/develop/client/js/upload-handler/form.upload.handler.js#L114)
```js
var fileId = getFileIdForIframeName(iframeName),
      response = handler._parseJsonResponse(message),
      uuid = response.uuid, // here
      onloadCallback;
```
I think it is not necessary~

rnicholus:
It is necessary

ICELI:
@rnicholus it‘s for security？

ICELI:
https://github.com/FineUploader/fine-uploader/blob/develop/client/js/upload-handler/form.upload.handler.js#L88
uuid  is already statemented

PR #1873, opened 2017-06-30
refactor: it is not necessary include the UUID of the response JSON #1872

ICELI:
- remove redundant code, `fileId` is unused.
- refactor callback usage, remove uuid condition.

windows 10, IE8,9

No test case before.


rnicholus:
This will break cross origin uploads in older browsers

ICELI:
Thank you for your reply.
I have test this in IE8,9... and i can't find some docs for this. Could you show some for me?

Issue #1888, opened 2017-07-24
Allow option to send chunking specific params as custom header params


thickl:
## Type of issue
- [ ] Bug report
- [x] Feature request
- [ ] Support request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

<details>
<summary>Feature Request</summary>

#### Feature request details 
{why is this feature important, not just for you, but for many others?}
For request of `traditional` uploads, there is an option to set params to be sent within the body (see https://docs.fineuploader.com/branch/master/api/options.html#request.paramsInBody). For an internal project I need the possibility to include the params for chunking requests as custom headers instead, in order to avoid numerous `pre-flight` requests for `cors`.

For this purpose I require to configure the options for params in a different way, for example by replacing the option `request.paramsInBody` with `request.paramsMode` and a new option -> `header | body | url`. Of course this approach would cause a breaking change in the core options API, so I am open for another solution.
Optionally for `request.header` there should be an option to set a prefix for header params (i.e. 'x-upload-'). This could be also set as 'request.headerParamPrefix`.

</details>

rnicholus:
Custome headers will not help you to avoid preflight requests. And with that I’m going to close this issue since that is the only obvious motivation for this proposal.

rnicholus:
This should be covered by #1917.

PR #1892, opened 2017-07-27
Allow option to send chunking specific params as custom header params…

thickl:
Allow option to send chunking specific params as custom header params (#1888)

For requests of traditional uploads, there is now a way to set params as custom headers.
There is a new option `request.paramsInBody` with `request.paramsMode` with the values -> `paramsInHeader | paramsInBody | paramsInUrl`. This causes a breaking change in the core options API, as the old option `paramsInBody: true | false` is removed.
Optionally for request.header there is an option to set a prefix for header params (i.e. 'x-upload-'). This can be set by setting a string value for 'request.headerParamPrefix`.

{example: Safari on iOS 9.1.0 and IE11 on Windows 8.1}
current Chrome / Firefox

{unit tests should accompany almost all PRs, unless the change is to documentation}
Yes, refactored existing tests and added new tests for custom headers.


rnicholus:
This should be covered by #1917.

PR #1933, opened 2017-10-25
Add serverside encryption headers as request params

nadeesha:
This allows us to specify custom encryption keys via `aws:kms`. It solves the problem that #1803 is trying to solve. With this, you can just do the following to get KMS keys working

```js
      request: {
        endpoint: "...",
        accessKey: "...",
        params: {
          "x-amz-server-side-encryption": "aws:kms",
          "x-amz-server-side-encryption-aws-kms-key-id": "JE9Yya8Rd6jcbZbb8ymFoU6prbpyyZDc"
        }
      },
```

Chrome. But this is unlikely to break any behaviour that is already not broken by custom params.

The current custom params implementation does not contain unit tests.


rnicholus:
This seems simple enough. I’ll try to merge it in soon.

Issue #1936, opened 2017-11-04
Double Thumbnail loading - Initial Filelist


blacktoby:
## Type of issue
- [x] Bug report
- [ ] Feature request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

<details>
<summary>Bug Report</summary>

#### Fine Uploader version
5.15

#### Browsers where the bug is reproducible
Chrome

#### Operating systems where the bug is reproducible
MacOS

#### Exact steps required to reproduce the issue
1. Get Filelist from Server with Thumbnail URLs 
2. Open Initial Filelist
3. In Chrome Debugger see that Thumbnails are loaded twice.

#### All relevant Fine Uploader-related code that I have written
Javascript for the FileUploader:
```
function openUploaderModal(element) {
    modal = new tingle.modal({
        footer: true,
        stickyFooter: false,
        closeLabel: "Close",
        beforeClose: function() {
            return true; // close the modal
        }
    });
    modal.setContent('<div id="uploader"></div>');
    
    var request = { endpoint: getCmsUrl()+'/api/cms/image-upload/uploads.json' };

    var uploader = new qq.FineUploader({
        element: document.getElementById("uploader"),
        request: request,
        cors: {
            expected: true,
            sendCredentials: true
        },
        session: {
            endpoint: getCmsUrl()+'/api/cms/image-upload/file-list'
        },
        deleteFile: {
            enabled: true,
            forceConfirm: true,
            endpoint: getCmsUrl()+'/api/cms/image-upload/delete'
        },
        callbacks: {
            onAllComplete: function(ids){
                var i;
                for(i = 0; i < ids.length; i++)
                {
                    $(".qq-file-id-"+ids[i]).on("click", function (e) {
                        if($(e.target).is("button")){
                            return;
                        }
                        imageSelected($(this), element);
                    })
                }
            },
            onSessionRequestComplete: function (response) {
                modal.open();
            },
            onComplete: function(id, name){
                console.log(id, name);
            }
        }
    });
}
```

#### Detailed explanation of the problem
When I open the initial Filelist with the Javascript above, all Thumbnails are loaded twice from the thumbnail urls. You can see that in the network section of the Chrome Debugger.
When there are a lot of pictures in the File list, the drawing of the thumbnails is very slow, so I debugged the issue and fixed it by doing this:
In fine-uploader.js:
```
updateThumbnail: function(id, thumbnailUrl, showWaitingImg, customResizeFunction) {
[...]

PR #1937, opened 2017-11-04
Double Thumbnail loading (#1936) Fix

blacktoby:
See Issue #1936 for more Infos

Changes related to #1936 

Chrome on MacOS

I have not written unit tests, because I dont have the complete overview over the code and do not have the time to working into it and setting up a proper environment to develop tests for this piece of code.
The fix I presented in #1936 is just a suggestion and is working in my situation. There should be someone other, who looks at it and test the change with a look at the full code.

### DIFF
--- a/client/js/templating.js
+++ b/client/js/templating.js
@@ -993,6 +993,7 @@ qq.Templating = function(spec) {
 
         updateThumbnail: function(id, thumbnailUrl, showWaitingImg, customResizeFunction) {
             if (!this.isHiddenForever(id)) {
+                if(showWaitingImg !== true) { return; }
                 thumbGenerationQueue.push({customResizeFunction: customResizeFunction, update: true, id: id, thumbnailUrl: thumbnailUrl, showWaitingImg: showWaitingImg});
                 !thumbnailQueueMonitorRunning && generateNextQueuedPreview();
             }


rnicholus:
Please explain your change in detail (the logic behind it, etc).

blacktoby:
I searched a way to avoid duplicate adding Thumbnails to the `thumbGenerationQueue`. 
I saw that the `updateThumbnail` is called two times:
One time in `_addToList`  and one time in `_maybeUpdateThumbnail` (part of `qq.uiPrivateApi`). 
`_addToList` is called when `_onSubmit` runs and `_maybeUpdateThumbnail` is called when `_markFileAsSuccessful` runs.
At this point I dont have a deeper look how fineuploader is working. So I just compared the parameters given when `updateThumbnail` is called. I saw that the first time `showWaitingImg` is true and the second time undefined.
I also observed, that the second call of the update Function is not necessary to show the thumbnail image. So I just excluded the second push to the queue with:
```
if(showWaitingImg !== true) { return; }
```
May be not the best way to handle this problem. For me it was the best way to get this problem solved, without deep dive into the code and change major procedures.

rnicholus:
Thanks for that. I’ll hopefully taje a closer look soon. The build will need to be fixed here as well, as one of the lint checks is failing

rnicholus:
After looking at this a bit closer, I'm (first) not really sure I understand the problem. And second I don't think the changes here are appropriate at all. I will say the call to `templating.updateThumbnail` in the `_maybeUpdateThumbnail` method of uploader.api.js is incorrect in that it fails to pass a boolean for the third parameter (which is `showWaitingImg`). But it's not clear that this is related to the issue you're having.

PR #1950, opened 2017-11-18
bloburi param in doc now matches code

dapug:
Minor edit to get the docs to match the code on the SAS request params.  HOWEVER, we might consider changing the code instead of, or in addition to this PR.  Seems kinda odd to have one variable with an underscore, and the others without:  `_method`, `bloburi`, `qqtimestamp`.


rnicholus:
Thanks for this. `_method` is a common convention used to “override” the HTTP request method.

PR #1964, opened 2018-01-21
Fix for https://github.com/FineUploader/fine-uploader/issues/1172

### LINKED ISSUE #1172: retry count does not reset with chunked uploads to S3
We are debugging an issue our client is seeing in Firefox with the S3 uploader where it sometimes fails with

```
[Fine Uploader 4.3.1] Received response status 400 with body: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>RequestTimeout</Code><Message>Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed.</Message>
```

The uploader has chunking enabled because the files being uploaded are typically 2 to 6 GB.

I have no idea as to the cause of that error (ideas?), but with the retry feature enabled, the failed chunk seems to successfully upload after 1 or 2 retries. The problem is the retry counter doesnt appear to reset to 0 if the chunk does upload successfully.

As it stands, it seems we need to set an arbitrary large number for maxAutoAttempts to accommodate a large number of retries on these files that are chunked in to over 1,000 pieces.

galvinhsiu:
Fix for the long standing issue on uploading using chunking,
 this usually kicked in when uploading very large files to S3 (1+
 gigabytes).  This patch alleviates the need to inflate the autoretry attempt number to be huge number, which may have ramifications on the end user.

The change is on a successful upload chunk, if autoretry is turned on and the upload is not cancelled or paused to reset the autoretry counter to 0 for the given upload handle.

Please refer to:
(https://github.com/FineUploader/fine-uploader/issues/1172).

Recommendations / comments are welcome.
{example: Safari on iOS 9.1.0 and IE11 on Windows 8.1}

Safari on Mac Sierra
< Firefox 50, > Firefox 50 on Mac Sierra
IE11 on Windows 10 / 8
Edge 14, Edge 15+ on Windows 10

Yes, see attached code.  Additional test added to fail out each chunk exactly once and then retry successfully, with the check on the attempt number = 1.

fixes #1172


test/unit/chunked-uploads.js
```diff
@@ -197,6 +200,86 @@ if (qqtest.canDownloadFileAsBlob) {
             });
         }
 
+        function testChunkedEveryFailureAndRecovery(done) {
+            assert.expect(6 + (expectedChunks * 17) + (expectedChunks * 6), done);
```
rnicholus: I'll admin I wish I hadn't added the ability to track asserts like this - it has become a maintenance mess. I ask you do one of two things here:

Either heavily comment this line to explain the significance of this number, or rewrite this test (possibly breaking it into smaller ones) such that assert counting is no longer needed.

client/js/uploader.basic.api.js
```diff
@@ -1574,6 +1575,14 @@
             //nothing to do in the base uploader
         },
 
+        _onUploadChunkSuccess: function(id, chunkData) {
+            var uploadData = this._uploadData.retrieve({id: id});
+
+            if (!this._preventRetries[id] && this._options.retry.enableAuto && uploadData.status !== qq.status.PAUSED) {
```
rnicholus: Why would the status ever be "paused" here?

rnicholus (changes requested):
Looks pretty good, but if you could address my two comments...

galvinhsiu:
Good feedback, ripped out the unused condition and slightly tweaked the unit test.

rnicholus (dismissed):

rnicholus:
Awesome, thanks! I'd like to get this merged in, and maybe you can do that. Let's continue our discussion on Twitter DM.

PR #1965, opened 2018-01-28
Adding a option to getParams and expandParams

seltix5:
currently the only function to set params ( "setParams" ) will replace the current data with new data, this is not always enough.
Adding "getParams" and "expandParams" will add the option to manipulate the params without erase them.
Without this options it is not possible to update multiple times the params from different sources.

i think this 2 simple functions are useful, what do you think?

### DIFF
--- a/docs/api/methods.jmd
+++ b/docs/api/methods.jmd
@@ -316,16 +316,18 @@ A `resizeInfo` object, which will be passed to the supplied function, contains t
 {{ api_method("getParams", "getParams (id)", "Get the parameters for an upload request. Pass in a file id to make the parameters specific to that file.",
 [
     {
-        "name": "params",
-        "type": "Object",
-        "description": "The parameters to include in the upload request."
-    },
-    {
         "name": "id",
         "type": "Integer",
         "description": "The file id."
     }
-], null) }}
+],
+[
+    {
+        "type": "Object",
+        "description": "Object containing parameters for the upload request."
+    }
+
+]) }}
 
 {{ api_method("getUuid", "getUuid (id)", "Returns the UUID of the item with the associated id.",
 [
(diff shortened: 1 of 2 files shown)


rnicholus (changes requested):
This does seem useful, but in order to merge this in, I’ll need two more updates:

- documentation
- _comprehensive_ unit tests that cover these new methods.

seltix5:
I do can update the documentation, but about the tests I dont know how it works or what I need to do :/

seltix5:
Docs updated. Can you help me with the tests?

rnicholus:
Sure - i'll try to make some time to write these soon.

docs/api/methods.jmd
```diff
@@ -313,6 +313,20 @@ A `resizeInfo` object, which will be passed to the supplied function, contains t
 
 ]) }}
 
+{{ api_method("getParams", "getParams (params[, id])", "Get the parameters for an upload request. Pass in a file id to make the parameters specific to that file.",
+[
+    {
+        "name": "params",
+        "type": "Object",
+        "description": "The parameters to include in the upload request."
```
SinghSukhdeep: @seltix5 this is incorrect as `getParams` method doesn't need this parameter. Can you please correct this?.
I'll be happy to write tests for this

singhjusraj (changes requested):

docs/api/methods.jmd
seltix5: done ^^

docs/api/methods.jmd
```diff
@@ -313,6 +313,20 @@ A `resizeInfo` object, which will be passed to the supplied function, contains t
 
 ]) }}
 
+{{ api_method("getParams", "getParams (id)", "Get the parameters for an upload request. Pass in a file id to make the parameters specific to that file.",
+[
+    {
```
SinghSukhdeep: @seltix5 this object should also be removed from here and return statement needs to return an object
Currently you have `null` there, it should be like:

```
[
    {
        "type": "Object",
        "description": "Object containing parameters for the upload request"
    }

]
```

Issue #1976, opened 2018-03-01
Reported directory path (qqpath) is wrong if file name occurs in one of the parent directory names


ogvolkov:
## Type of issue
- [x] Bug report
- [ ] Feature request

## Uploader type
- [x] Traditional
- [ ] S3
- [ ] Azure

<details>
<summary>Bug Report</summary>

#### Fine Uploader version
5.15.6

#### Browsers where the bug is reproducible
Chrome 64.0.3282.167, FF 58.0.2, Edge 41.16299.248.0

#### Operating systems where the bug is reproducible
Windows 10

#### Exact steps required to reproduce the issue
1. Setup Fine Uploader UI with dragAndDrop.reportDirectoryPaths set to true.
2. Prepare a directory with files on the local filesystem where one of the file names appears in a directory name. For example, create directory named "data" containing a file named "data".
3. Drop the directory onto Fine Uploader.
4. Observe that qqpath is incorrect or missing from the data passed to the backend. In the data/data example, qqpath parameter will not be present in the request altogether.

#### All relevant Fine Uploader-related code that you have written
```
var uploader = new qq.FineUploader({
            element: document.getElementById("uploader"),
            multiple: true,
             dragAndDrop: {
                reportDirectoryPaths: true
            }
        })

```
</details>

rnicholus:
what does "incorrect" mean? Looking at the code, i'm not seeing any evidence of an issue. Can you provide more concrete details?

rnicholus:
If you can update the issue with all of your Fine Uploader config code _and_ the full request/response headers+payload as well as any console.log statements, that would be helpful.

ogvolkov:
There is a pull request which fixes it already.

But to clarify, if I upload a folder "data" which contains a folder "sub" which contains a file "ub", the resulting payload is
------WebKitFormBoundaryxwMePIf55gNLuxW1
Content-Disposition: form-data; name="qqpath"

data/s
------WebKitFormBoundaryxwMePIf55gNLuxW1
Content-Disposition: form-data; name="qquuid"

8bbf2d77-7f76-457d-97f4-a327afb673fa
------WebKitFormBoundaryxwMePIf55gNLuxW1
Content-Disposition: form-data; name="qqfilename"

ub
 etc.

The reason is the code tries to remove the file name from the path by find a first occurrence via indexOf, so "/data/sub/ub" without "ub" becomes "/data/s", which clearly incorrect

PR #1977, opened 2018-03-01
Fix reported path when file name also occurs in parent directory name(s)

ogvolkov:
Fixes reported path (qqpath) in case when the file name in question appears in one of its parent directory names, see #1976. The original code used indexOf so if the first occurrence was not a file name itself, it stripped way too much from the path.

Chrome 64.0.3282.167, FF 58.0.2, Edge 41.16299.248.0

Unit tests are present, both for normal case where I used the value the current code returns as a reference, and for the case when directory name is the same as the file name.

### DIFF
--- a/client/js/dnd.js
+++ b/client/js/dnd.js
@@ -34,19 +34,7 @@ qq.DragAndDrop = function(o) {
 
         if (entry.isFile) {
             entry.file(function(file) {
-                var name = entry.name,
-                    fullPath = entry.fullPath,
-                    indexOfNameInFullPath = fullPath.indexOf(name);
-
-                // remove file name from full path string
-                fullPath = fullPath.substr(0, indexOfNameInFullPath);
-
-                // remove leading slash in full path string
-                if (fullPath.charAt(0) === "/") {
-                    fullPath = fullPath.substr(1);
-                }
-
-                file.qqPath = fullPath;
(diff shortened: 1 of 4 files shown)


client/js/dnd.js
```diff
@@ -85,6 +73,22 @@ qq.DragAndDrop = function(o) {
...
+    function extractDirectoryPath(entry) {
+        var name = entry.name,
+            fullPath = entry.fullPath,
+            indexOfNameInFullPath = fullPath.lastIndexOf(name);
+
+        // remove file name from full path string
+        fullPath = fullPath.substr(0, indexOfNameInFullPath);
+
+        // remove leading slash in full path string
+        if (fullPath.charAt(0) === "/") {
```
rnicholus: This may be considered a breaking change, which I want to avoid at all costs. Was this required to fix the underlying issue?

rnicholus:
just to be clear, the only change was `fullPath.indexOf(name)` -> `fullPath.lastIndexOf(name)`?

ogvolkov:
Yes, the functional change is only that.
However to have it tested I've extracted it as a separate function, which is also a bit cleaner imo.

rnicholus:
thanks for your work on this! I'm going to release this as 5.15.7 shortly

PR #1529, opened 2016-02-16
Remove eval() fallback for JSON parsing.

KevinGrandon:
(Now branched form develop)

Currently eval() is only used for IE7 and prevents sites from implementing content security policies. Since IE7 is a legacy browser, I suggest releasing a fine-uploader version without IE7 support, allowing those who wish to keep legacy support the ability to use an older version of this library.

Refs #1438

PR #1983, opened 2018-03-23
refactor: use native JSON.parse, drops IE7 support

bradleyayers:
Currently `eval()` is only used for IE7 and prevents sites from implementing content security policies.

Refs #1438, #1529

### DIFF
--- a/client/js/s3/request-signer.js
+++ b/client/js/s3/request-signer.js
@@ -272,7 +272,7 @@ qq.s3.RequestSigner = function(o) {
         // Attempt to parse what we would expect to be a JSON response
         if (responseJson) {
             try {
-                response = qq.parseJson(responseJson);
+                response = JSON.parse(responseJson);
             }
             catch (error) {
                 options.log("Error attempting to parse signature response: " + error, "error");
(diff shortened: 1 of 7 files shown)


rnicholus:
This is not true, and `eval` will only be used if `JSON.parse` is not available, which will only occur on browsers that don't even support CSPs.

rnicholus:
Additionally, this is a breaking change, which is not to be taken lightly. I fully intend to remove support for anything older than IE11, but in a major version update - 6.0.

rnicholus:
Well, Fine Uploader follows the semantic versioning standard, so breaking changes will only ever (knowingly) occur as part of a major version release. There's currently a release/6.0.0 branch, but it's quite stale and I haven't had time to work on 6.0.0 for a while. I spent the last 4+ months on 5.16.0, which was a huge release that came out a week or so ago - https://github.com/FineUploader/fine-uploader/releases/tag/5.16.0

bradleyayers:
I saw the 5.16.0 release and it looks like a lot of really nice improvements! Congratulations on getting that out!

bradleyayers:
I've updated the base branch to `release/6.0.0`, are you happy to merge this to there?

rnicholus:
Sounds good to me, thanks

PR #1989, opened 2018-03-27
fix: TypeScript declaration for lib/core/s3; add retry option

### LINKED ISSUE #1988: Support TypeScript definitions for fine-uploader/lib/core/*
## Type of issue
- [ ] Bug report
- [x] Feature request

## Uploader type
- [x] Traditional
- [x] S3
- [x] Azure

<details>
<summary>Feature Request</summary>

#### Feature request details 

I'm using TypeScript and I want to use the `fine-uploader/lib/core/s3` module, but there are currently no types declared for that module.
</details>

bradleyayers:

- Adds TypeScript declaration for `lib/core/s3`.
- Adds `retry` option to `FineUploaderBasic`.

Resolves #1988.

Extra `package.json` files have been added to `commonJs` to associate types with nested modules, this follows the pattern used by date-fns.

n/a

`fine-uploader.test.ts` has been extended.


rnicholus:
@SinghSukhdeep can you take a peek at this?

SinghSukhdeep:
@rnicholus Sure I'll take a look

singhjusraj (commented):
I will take another deeper look soon. Following is what I've noticed so far.

client/typescript/fine-uploader.d.ts
```diff
@@ -3062,10 +3066,12 @@ declare module "fine-uploader/lib/azure" {
              * RequestOptions
              */
             request?: AzureRequestOptions;
-
+            /**
+             * Retry options
+             */
+            retry?: UIRetryOptions & RetryOptions;
```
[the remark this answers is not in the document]
bradleyayers: I had the same thought, but actually this is required to be explicitly merged as TypeScript does not allow implicit merging of incompatible types via multiple inheritance (`xxx extends Foo, Bar`).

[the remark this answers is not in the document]
SinghSukhdeep: Yes I have encountered that before. Then here we would only need to specify `UIRetryOptions` and not both as `UIRetryOptions` already extends the `RetryOptions`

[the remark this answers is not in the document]
bradleyayers: Ah I understand now! I've updated to be just `UIRetryOptions` — thanks

client/commonJs/traditional.js
```diff
@@ -1,3 +0,0 @@
-"use strict";
-
-module.exports = require("../fine-uploader/fine-uploader");
```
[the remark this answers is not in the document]
bradleyayers: I didn't intend to make any changes to the way people import modules. They should still be able to import `fine-uploader/lib/traditional`. My understanding was that `lib/traditional.js` and `lib/traditional/index.js` can both be imported via `lib/traditional`.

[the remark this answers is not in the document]
SinghSukhdeep: It should be fine according to my understanding too.
But if we are referencing `fine-uploader/lib/traditional.js` directly, as it is needed for Rollup config, that would need to be changed to `fine-uploader/lib/traditional/index.js`, which seems like a reasonable small change that only affects people who utilize Rollup, primarily Angular users.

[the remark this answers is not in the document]
bradleyayers: Oh I see, I hadn't noticed those before. Yes I'll update those in this pull request. Good catch!

SinghSukhdeep:
We should also update the azure module similarly as that is also missing the core module declaration

SinghSukhdeep:
@bradleyayers Just wanted to check your thoughts on possibility of `'fine-uploader/lib/all'` in TS. This is missing and I can't think of a way to have this since we don't have the `qq` namespace here

bradleyayers:
@SinghSukhdeep I've pushed up a couple of commits attempting to add `lib/all` and `lib/core/all`. To be honest I'm not *exactly* sure what exists in those distributions, so if you can double check I got it right that would be great. It's times like these I hope having an ES module distribution would remove the need for these separate distributions. I'd really like to have the whole code base migrated to TS to make these things easier.

bradleyayers:
@SinghSukhdeep let me know if you're aware of any other issues, else I think this is good to go.

client/typescript/fine-uploader.d.ts
```diff
@@ -2691,6 +2693,14 @@ declare module "fine-uploader" {
 
 }
 
+declare module "fine-uploader/lib/all" {
+    import { azure } from 'fine-uploader/lib/azure';
+    export * from 'fine-uploader/lib/core';
```
SinghSukhdeep: @bradleyayers I think this should be `export * from 'fine-uploader'` instead of core

SinghSukhdeep:
Hey sorry I was away for couple days. I think it all looks great except for one line I mentioned above

client/typescript/fine-uploader.d.ts
bradleyayers: Ah interesting. So do you think `azure` and `s3` should be exported from `fine-uploader`?

SinghSukhdeep: `azure` and `s3` is fine here, It's just that we are exporting the core `fine-uploader` from here, it should be the UI module which is `fine-uploader` and not `fine-uploader/lib/core`

bradleyayers: Makes sense, and do you think `azure` and `s3` should *also* be exported from `fine-uploader`? I thought they were accessible via `qq.s3.*` etc

SinghSukhdeep: `azure` and `s3` should be exported from the `fine-uploader/lib/all` module but not from the `fine-uploader` module.
And `FineUploaderBasic` isn't exported from `fine-uploader`.
That must be imported from `fine-uploader/lib/core`

SinghSukhdeep: I think since we are not exporting `FineUploaderBasic` from `fine-uploader` module, we shouldn't export the same from `fine-uploader/lib/azure` and `fine-uploader/lib/s3` as well.
The core modules should only be exported from the `fine-uploader/lib/core`, `fine-uploader/lib/core/azure` and `fine-uploader/lib/core/s3`

That's what makes more sense to me.

bradleyayers: Okay, I've removed the FineUploaderBasic export from the non-core lib modules, and switched to `export *` as suggested.

SinghSukhdeep: Great, I think we're good to go

singhjusraj (approved):

bradleyayers:
@rnicholus this should be good to go.

rnicholus:
Everything looks fine here, but I'm not sure I understand the documentation changes. You shouldn't have to explicitly reference an index.js file when importing a module, unless I'm missing something. If that is the case here, then this would be a very significant breaking change, and we'll have to figure out how to update the code to prevent that.

bradleyayers:
Fair call, this might be better suited for `release/6.0.0`?

rnicholus:
Is it really necessary to reference index.js directly? I’ve never had to do that myself.

rnicholus:
So, does this there are no breaking changes then?

bradleyayers:
> So, does this there are no breaking changes then?

"Breaking change" is a subjective term, so I think this boils down to being a judgment call. I'm happy to call this not a breaking change, even though if people are relying on exact file locations for custom build configurations they'll need to make some changes when upgrading. If you'd rather err on the side of calling that a breaking change, then I can target this to `release/6.0.0`. 

Without an unambiguous definition for breaking change this decision just needs to be a judgement call made by a maintainer.

rnicholus:
If any user needs to adjust their code to upgrade, then the version contains breaking changes, and that must always be avoided in minor or patch releases, per semver spec. I’ll take a closer look at these changes soon to better determine if the documentation changes are necessary and if any breaking changes are present or necessary.

SinghSukhdeep:
Just to weigh in, we only need to specify file extensions for module loaders like SystemJS (not sure about Webpack) and module bundlers like Rollup (again not sure about Webpack).
As far as normal import export is concerned, having or specifying `index.js` in the docs doesn't matter.

So only a slight modification is required in the bundler/loader configs.

And in our `package.json`
`"main": "lib/traditional.js"` would also needs to be changed to `"main": "lib/traditional/index.js"`

bradleyayers:
@SinghSukhdeep does `pkg.main` support having the file extension omitted?

rnicholus:
I think i'm ok with this provided the documentation changes are reverted. But I do want to do a bit of manual testing myself first. Thanks both of you for all of your hard work.

bradleyayers:
Reverting the documentation changes would also require reverting the code changes that move `lib/traditional.js` to `lib/traditional/index.js`, else the docs won't match up with the file structure.

rnicholus:
I’m personally fine with just leaving the docs as-is (before any changes). It’s common to use the index.js pattern (I’m using it in several closed and open source projects) and I’ve never seen docs explicitly reference the index file.

rnicholus:
I’m fairly certain everything will just work without making any code changes in webpack environments. I’ll have to confirm though. But if indeed this does break for users of rollup, then without reverting the breaking code changes, this will have to sit and wait to go out with 6.0

bradleyayers:
> To ensure we don't break backwards compatibility, can't we just keep the traditional.js, s3.js, etc files?

Yeah I considered that, but I wasn't sure if the module resolution spec includes whether to try directory or file first when resolving. It seemed like a risky approach that could be broken in the future. It's probably worth checking though.

rnicholus:
Well, if both /traditional/traditional.js and traditional/index.js point to the same location, there shouldn't be an issue, unless I'm missing something. We could eliminate traditional.js in 6.0, but this allows us to ensure there are no breaks in 5.x.

SinghSukhdeep:
Hi @bradleyayers are you still pursuing this PR. We did a lot of work on this and it would be great if we could merge this.

bradleyayers:
@SinghSukhdeep I'm not actively working on it, but I am still using my FineUploader fork in production (it includes this change + using native promises), so I'm still invested in seeing this succeed.

SinghSukhdeep:
Awesome!
I will look into this again and see what is missing and hopefully we can get this through

PR #1998, opened 2018-04-23
docs: fix Amazon S3 v4 signature guide

bradleyayers:
Fixes an error in the docs.


rnicholus:
Good catch, Bradley. Thanks for your work!

PR #2014, opened 2018-06-05
Clarified that callbacks are FineUploader option.

Martinsos:
Thanks for this awesome library!
I have been using it on a project to perform upload to Azure from browser with completely custom UI.

I had hard time figuring out where to define `callbacks` while reading docs, as it is never explicitly stated that `callbacks` is a FineUploader option. I ended up trying different things and at the end it worked if I would provide as FineUploader option. Therefore, I thought it would be useful to explicitly state that in docs, and that is what this PR does.

None, because I did not manage to run building of documentation. I inspected Makefile and tried downloading docfu manually but realized that should probably not be the way to build it and stopped there, because I don't have a good idea how to build it.
Therefore, I am not sure how the PR is going to render. If you could point me toward the way to do this I will, or you can try it out and fix any errors.

No, because I changed only documentation.

### DIFF
--- a/docs/api/events.jmd
+++ b/docs/api/events.jmd
@@ -45,6 +45,8 @@ callbacks: {
 }
 ```
 
+where `callbacks` is specified as a FineUploader option.
+
 {% endmarkdown %}
 <div class="events-all">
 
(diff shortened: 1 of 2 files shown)


Martinsos:
Awesome, thanks!
Question for the future, if I change a piece of docs, is there any easy way to render it and check what it looks like?

PR #2021, opened 2018-07-02
S3: enable overriding x-amz-date from signing server

ViliamKopecky:

AWS S3 SDK using PostObjectV4 (https://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.S3.PostObjectV4.html) doesnt let you set custom date for policy, and sometimes the dates are different about a second, so the better way for me is setting the date from server.

Maybe even better way would be setting the whole `awsParams` from server.

macos chrome

I didnt find existing matching test for this.


donbobka:
+1 to enable overriding of any field (not only date)?

PR #2022, opened 2018-07-03
Fix an issue where onAllComplete can be called multiple times per addFile() call

AndrewJDR:
_getNotFinished, which is used to determine when to fire the onAllComplete callback, wasn't accounting for files in the UPLOAD_FINALIZING state. For smaller files, this could cause onAllComplete to be called multiple times per set of files passed to addFiles().

Note, the expected result is for onAllComplete() to be called only once per call to addFiles().

### DIFF
--- a/client/js/uploader.basic.api.js
+++ b/client/js/uploader.basic.api.js
@@ -982,6 +982,7 @@
             return this._uploadData.retrieve({
                 status: [
                     qq.status.UPLOADING,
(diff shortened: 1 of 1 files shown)


oles-bolotniuk:
+1 Please merge

This is pretty much critical, because as `finalizing` process of some files may take time - `onAllComplete` gets called twice:

1. First time when all uploads are done
2. Second time when finalizing of the last file is done

Issue #1882, opened 2017-07-07
TypeError: target.onload is not a function


hirenalken:
## Type of issue
Bug report

## Uploader type
 S3

#### Exact steps required to reproduce the issue
For example:
1. Using s3.fine-uploader.core.js
2. Creating own UI from scratch and using **img** element as container for image preview

#### Error

```
core.es5.js:1085 ERROR TypeError: target.onload is not a function
    at s3.fine-uploader.core.js:4151
    at qq.Promise.then (s3.fine-uploader.core.js:709)
    at s3.fine-uploader.core.js:4148
    at MegaPixImage.render (s3.fine-uploader.core.js:4153)
    at Array.<anonymous> (s3.fine-uploader.core.js:4119)
    at s3.fine-uploader.core.js:4104
    at ZoneDelegate.invokeTask (zone.js:424)
    at Object.onInvokeTask (core.es5.js:4136)
    at ZoneDelegate.invokeTask (zone.js:423)
    at Zone.runTask (zone.js:191)
```

rnicholus:
Please show your code

hirenalken:
I am using this with angular 2. insalled fineuploader via npm. 

To import 
`import qq from 'fine-uploader/lib/core/s3';`

```
ngOnInit() {
    this.setAlbumModal();
    let browsedImages: BrowsedImage[] = [];
    let instance = this;

 this.uploader = new qq.s3.FineUploaderBasic({
      button: document.getElementById('upload_image'),
      debug: false,
      autoUpload: false,
      multiple: true,
      validation: {
        allowedExtensions: ['jpeg', 'jpg', 'png', 'gif'],
        sizeLimit: 5120000 // 50 kB = 50 * 1024 bytes
      },
      region: 'us-west-2',
      request: {
        endpoint: 'https://xxxxx.s3.amazonaws.com/',
        accessKey: 'XXXXXXXXXXXX'
      },
      signature: {
        endpoint: 'http://xxxxxxxxxxxxxxx/s3_signature/',
        // version: 4,
      },
      uploadSuccess: {
           endpoint: 'http://localhost:3000?success=true'
       },    
      cors: {
        expected: true,
        sendCredentials: true
      },
      callbacks: {
        onSubmit: function (id, fileName) {
          console.log("ohh wow");
          instance.browsedImages.push({
            id: id,
            fileName: fileName
          });

          console.log(instance.browsedImages);
          // passing data to child component for display. which is a modal component
[...]

PR #2056, opened 2018-10-05
Call target.onload() only when defined

Javarome:
This PR checks that the `onload` function is defined before calling it. This removes the `Uncaught TypeError: target.onload is not a function` console error while img preview.

Chrome 69.0.3497.100 on MacOS Sierra 10.13.6

No, because the modified code is likely already tested.

### DIFF
--- a/client/js/image-support/megapix-image.js
+++ b/client/js/image-support/megapix-image.js
@@ -383,7 +383,7 @@
                 renderImageToDataURL(self.srcImage, self.blob, opt, doSquash)
                     .then(function(dataUri) {
                         target.src = dataUri;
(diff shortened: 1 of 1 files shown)


Javarome:
Hi @SinghSukhdeep 

I encountered it many times in my production code when using `` for preview. I am unable to share it here, but another user has reported here with some sample code : https://github.com/FineUploader/fine-uploader/issues/1882
(a PR was rejected about it because it prevented calling `onload` when defined, whereas this current PR will avoid calling it only when undefined).

SinghSukhdeep:
Okay sounds good. Will merge this and release with few other pending PRs

